PR #14, opened 2018-11-29
Support new severeties

yahavi:


plugin/src/main/java/org/jfrog/idea/xray/persistency/types/Issue.java
```diff
@@ -14,8 +14,6 @@
     public String description;
     public String issueType = "N/A";
     public String provider;
-    @Deprecated
-    public String sevirity;
```
eyalbe4: Are you sure we can remove this deprecated property? People may still be using an old version the plugin, with this property in the cache, and upgrading to this version will cause issues.
What do you think?

plugin/src/main/java/org/jfrog/idea/xray/persistency/types/Severity.java
```diff
@@ -4,5 +4,59 @@
...
+        return this.ordinal > other.getOrdinal();
+    }
+
+    public static Severity fromString(String inputSeverity) {
+        for (Severity severity : Severity.values()) {
+            if (severity.getSeverityName().equals(inputSeverity)) {
+                return severity;
+            }
+        }
+        // Backward compatibility
```
eyalbe4: Question: 
Why is this switch case for the deprecated values needed? Wouldn't the above loop cover the deprecated values as well?

eyalbe4:
Looks good @yahavi 
I did have two implementation questions (inline). Let me know what you think.
Thanks!

plugin/src/main/java/org/jfrog/idea/xray/persistency/types/Issue.java
yahavi: I thought its unlikely that people are still using the version from 2017 with Intellij's auto updates. I'll revert it to be on the safe side.

plugin/src/main/java/org/jfrog/idea/xray/persistency/types/Severity.java
yahavi: That wouldn't work because the new names of `Minor`, `Major` and `Critical` are `Low`, `Medium` and `High`.

PR #23, opened 2019-06-13
Fix tree insertion in the UI

yahavi:


src/test/java/com/jfrog/ide/idea/ui/BaseTreeTest.java
```diff
@@ -0,0 +1,98 @@
...
+        baseTree.reset();
+        Assert.assertNull(baseTree.getModel());
+
+        // Append single project again and check results
+        baseTree.appendProject(dependenciesTreeA);
+        Assert.assertEquals(baseTree.getModel().getRoot(), dependenciesTreeA);
+    }
+
+    @Test
+    void twoProjectsTest() {
```
eyalbe4: I suggest adding a description comment for each of the **Assert**s in this method. 
This should make it easier to understand.
Same suggestion for the *threeProjectsTest* method.

PR #35, opened 2019-12-17
Read certificates from Idea certificates manager

yahavi:


yahavi:
https://github.com/jfrog/jfrog-idea-plugin/issues/34

PR #38, opened 2019-12-23
Add dependency information in project descriptor

yahavi:
1. Add "Show in dependencies tree" button.
2. Add "Licenses" and "Top issue" labels.

Bumped the minimal version to 2018.1 in order to fully support these features in Maven.

### DIFF
--- a/src/main/java/com/jfrog/ide/idea/utils/Utils.java
+++ b/src/main/java/com/jfrog/ide/idea/utils/Utils.java
@@ -2,6 +2,9 @@ package com.jfrog.ide.idea.utils;
 
 import com.google.common.base.Objects;
 import com.intellij.openapi.project.Project;
+import com.intellij.openapi.wm.ToolWindow;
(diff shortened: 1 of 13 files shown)


src/main/java/com/jfrog/ide/idea/utils/Utils.java
```diff
@@ -39,6 +42,12 @@ public static boolean areRootNodesEqual(DependenciesTree lhs, DependenciesTree r
                 StringUtils.equals(lhsGeneralInfo.getPath(), rhsGeneralInfo.getPath());
     }
 
+    public static boolean gavStringEqual(GeneralInfo lhs, GeneralInfo rhs) {
```
eyalbe4: Can we replace this method with an equals method in the GeneralInfo class?

yahavi: No because these are only 3 out of 7 class variables of the General info. This is a private use case of general info comparison.

PR #43, opened 2019-12-28
Add inspection tests

yahavi:
Add inspection tests and fix some minor issues


src/main/java/com/jfrog/ide/idea/inspections/AbstractInspection.java
```diff
@@ -210,6 +216,54 @@ DependenciesTree getRootDependenciesTree(PsiElement element) {
...
+                .map(DependenciesTree::getGeneralInfo)
+                .filter(Objects::nonNull)
+                .filter(grandChildGeneralInfo -> compareGeneralInfos(generalInfo, grandChildGeneralInfo))
+                .findAny()
+                .ifPresent(grandChildGeneralInfo -> modules.add(child)));
+        return modules;
+    }
+
+    /**
+     * Return true iff the dependency stated in the General info is a module in the project.
```
eyalbe4: iff --> if

src/main/java/com/jfrog/ide/idea/inspections/AbstractInspection.java
```diff
@@ -210,6 +216,54 @@ DependenciesTree getRootDependenciesTree(PsiElement element) {
...
+     *
+     * @param node        - The project node
+     * @param generalInfo - General info of the dependency
+     * @return true iff the dependency stated in the General info is a module in the project
```
eyalbe4: iff --> if

src/main/java/com/jfrog/ide/idea/inspections/AbstractInspection.java
yahavi: iff == if and only if

yahavi:
I tried to use the resources dir, but the test framework persists it should be under testData. I tried and It didn't work under resources. Please take look at their documentation under
[CodeInsightTestFixture::configureByFile](https://github.com/JetBrains/intellij-community/blob/master/platform/testFramework/src/com/intellij/testFramework/fixtures/CodeInsightTestFixture.java#L126).

PR #47, opened 2020-02-05
Use IntelliJ IDE PasswordSafe to store credentials

barbelity:


src/main/java/com/jfrog/ide/idea/configuration/GlobalSettings.java
```diff
@@ -52,11 +52,23 @@ public XrayServerConfigImpl getXrayConfig() {
...
     public void setXrayConfig(XrayServerConfigImpl xrayConfig) {
+        this.xrayConfig.setUrl(xrayConfig.getUrl());
+        this.xrayConfig.setUsername(xrayConfig.getUsername());
+        this.xrayConfig.setCredentials(xrayConfig.getCredentialsFromPasswordSafe());
+        this.xrayConfig.setExcludedPaths(xrayConfig.getExcludedPaths());
+    }
+
+    public void updateConfig(XrayServerConfigImpl xrayConfig) {
+        if (this.xrayConfig.getUrl() != null && !this.xrayConfig.getUrl().equals(xrayConfig.getUrl())) {
+            this.xrayConfig.removeCredentialsFromPasswordSafe();
```
yahavi: At first glance, it looks like updateConfig removes the credentials from the passwordSafe and set them again. Are you sure the deletion is necessary?

src/main/java/com/jfrog/ide/idea/configuration/GlobalSettings.java
```diff
@@ -52,11 +52,23 @@ public XrayServerConfigImpl getXrayConfig() {
...
+        }
         this.xrayConfig.setUrl(xrayConfig.getUrl());
         this.xrayConfig.setUsername(xrayConfig.getUsername());
         this.xrayConfig.setPassword(xrayConfig.getPassword());
         this.xrayConfig.setExcludedPaths(xrayConfig.getExcludedPaths());
+        this.xrayConfig.addCredentialsToPasswordSafe();
     }
```
yahavi: This 2 methods have common code. I suggest to reuse common code.
Please also add Javadoc above each one and explain the purpose of each one.

src/main/java/com/jfrog/ide/idea/configuration/GlobalSettings.java
barbelity: IMO it is necessary.
If a certain URL was previously saved to the password safe, and then the URL changes -> I think we should remove the old one.

barbelity:
I was using an incorrect API for creating the CredentialsAttributes object.
It should work now after restart as well.

PR #50, opened 2020-02-27
Improve inspection in Gradle

yahavi:
Inspect "Multi project - Single module":
* Project 1 -
  * Dependency 1
  * Dependency 2
* Project 2
* Project 3

This fix make sure that dependencies 1 and 2 will be inspected for licenses and issues in `build.gradle`. The regular dependencies tree is not affected from this.


yahavi:
@barbelity you are right!
+ Changed `getModulesFromTree` to `collectModules`
+ Put both `collectModules`s together
+ Added explanation the distinguish between npm and Go vs Maven and Gradle.

Please take a look 👀

PR #57, opened 2020-04-12
Initialize connection details from environment-variables

barbelity:


src/main/java/com/jfrog/ide/idea/configuration/XrayServerConfigImpl.java
```diff
@@ -197,10 +199,30 @@ void setPassword(String password) {
...
         setUsername(credentials.getUserName());
         setPassword(credentials.getPasswordAsString());
     }
 
+    /**
+     * Initialize connection details from environment variables.
+     * Minimal required details are url and username.
+     */
+    public void initConnectionDetailsFromEnv() {
+        String urlEnv = System.getenv(URL_ENV);
```
yahavi, line 214: This approach should work on 100% of the cases. See explanation above [getEnvironmentMap()](https://github.com/JetBrains/intellij-community/blob/webstorm/193.5233.13/platform/util/src/com/intellij/util/EnvironmentUtil.java#L128).
```suggestion
        String urlEnv = EnvironmentUtil.getValue(URL_ENV);
```

src/main/java/com/jfrog/ide/idea/configuration/XrayServerConfigImpl.java
```diff
@@ -197,10 +199,30 @@ void setPassword(String password) {
...
+        String usernameEnv = System.getenv(USERNAME_ENV);
```
yahavi, line 215:
```suggestion
        String usernameEnv = EnvironmentUtil.getValue(USERNAME_ENV);
```

src/main/java/com/jfrog/ide/idea/configuration/XrayServerConfigImpl.java
```diff
@@ -197,10 +199,30 @@ void setPassword(String password) {
...
+        String passwordEnv = System.getenv(PASSWORD_ENV);
```
yahavi, line 216:
```suggestion
        String passwordEnv = EnvironmentUtil.getValue(PASSWORD_ENV);
```

src/main/java/com/jfrog/ide/idea/configuration/XrayServerConfigImpl.java
```diff
@@ -197,10 +199,30 @@ void setPassword(String password) {
...
+        if (StringUtils.isBlank(urlEnv) || StringUtils.isBlank(usernameEnv)) {
+            return;
+        }
+        setUrl(urlEnv);
+        setUsername(usernameEnv);
+        setPassword(passwordEnv);
+        addCredentialsToPasswordSafe();
```
yahavi, line 223: I'm not sure that it is necessary in this case. We won't want to bother the user with password prompts when it is not really necessary.
```suggestion
```

barbelity: We decided to read the config-details once, and store them for the user.
Then, in future uses if the configuration is already set, we are not reading env-vars again.

PR #58, opened 2020-04-16
Persist issues and licenses filters for each project

barbelity:


src/main/java/com/jfrog/ide/idea/ui/filters/FilterManagerService.java
```diff
@@ -1,16 +1,79 @@
...
+        Map<License, Boolean> selectedLicences = super.getSelectedLicenses();
+        state.selectedLicences = new HashMap<>();
+        for (License license : selectedLicences.keySet()) {
+            state.selectedLicences.put(license.getName(), selectedLicences.get(license));
+        }
+
+        return state;
+    }
+
+    public void loadState(FiltersState state) {
```
yahavi, line 72:
```suggestion
    @Override
    public void loadState(FiltersState state) {
```

src/main/java/com/jfrog/ide/idea/ui/filters/FilterManagerService.java
```diff
@@ -1,16 +1,79 @@
...
+        if (state.selectedLicences != null) {
+            for (License license : selectedLicenses.keySet()) {
+                if (state.selectedLicences.containsKey(license.getName())) {
+                    selectedLicenses.put(license, state.selectedLicences.get(license.getName()));
+                }
+            }
+            state.selectedLicences = null;
+            // Update licenses tree with applied filters.
+            MessageBus messageBus = ApplicationManager.getApplication().getMessageBus();
+            messageBus.syncPublisher(ApplicationEvents.ON_SCAN_FILTER_LICENSES_CHANGE).update();
```
yahavi, line 47: This is a bit heavy operation. Shouldn't we do it only if there are unselected licenses? 🤔

src/main/java/com/jfrog/ide/idea/ui/filters/FilterManagerService.java
```diff
@@ -1,16 +1,79 @@
...
+            state.selectedLicences.put(license.getName(), selectedLicences.get(license));
+        }
+
+        return state;
+    }
+
+    public void loadState(FiltersState state) {
+        if (state == null) {
+            return;
+        }
```
yahavi, line 75:
```suggestion
```
I think the state object can't be null, because Intellij should call it only if there is a state to load. This is described also [here](https://github.com/JetBrains/intellij-community/blob/webstorm/193.5233.13/platform/projectModel-api/src/com/intellij/openapi/components/PersistentStateComponent.java#L40).

src/main/java/com/jfrog/ide/idea/ui/filters/FilterManagerService.java
yahavi, line 72:
```suggestion
    public void loadState(@NotNull FiltersState state) {
```

src/main/java/com/jfrog/ide/idea/ui/filters/FilterManagerService.java
barbelity: As this code runs only once at project load, I didn't add such check.
Do you think it worths it?

yahavi: Yes, because Intellij works hard during start.
```suggestion
            if (selectedLicenses.containsValue(false)) {
                MessageBus messageBus = ApplicationManager.getApplication().getMessageBus();
                messageBus.syncPublisher(ApplicationEvents.ON_SCAN_FILTER_LICENSES_CHANGE).update();
            }
```

barbelity: 👍

barbelity:
> Looks good! One thing that I think may cause confusion is:
> User that set a filter for license/severity and forget that he did it. I think we should show somehow that filters are applied.

We will add this improvement to our roadmap.

Issue #55, opened 2020-04-01
Attempting to install JFrog Plugin


kutran:
I have Android Studio 3.6.2 and trying to install JFrog Idea Plugin 1.5.0.
- first of all, can't find the plugin in the Marketplace
- when installing by disk, it gives the following message:
   > Plugin Error
   > Problems found loading plugins:
  > Plugin "JFrog" was not loaded: required plugin "org.jetbrains.idea.maven"

eyalbe4:
@kutran,
I'm not sure the JFrog IDEA Plugin is compatible with Android Studio. That's probably the reason why you couldn't find it on the Marketplace and got this error when attempting to install it from disk.

kutran:
Isn't Android Studio just a modified version of IntelliJ IDE?

eyalbe4:
It may be so @kutran, but the plugin has been tested only with IntelliJ IDEA. Looking forward,  we do want to support Android Studio, either by this plugin or a new one.

a1flecke:
@eyalbe4   Is this the best issue to follow for updates on Android Studio support for this plugin?  We would like to have XRay support in Android Studio as well.

eyalbe4:
Sure @a1flecke,
Support for Android Studio is part of our roadmap. We'll update this issue when we have more details to share.

OwenMoriarty-JNJ:
Any update on the  support for Android Studio please?

yahavi:
@OwenMoriarty-JNJ,
We are working on adding Android Studio and Gradle-Kotlin DSL support.
Here are the relevant pull requests:
https://github.com/jfrog/ide-plugins-common/pull/36
https://github.com/jfrog/jfrog-idea-plugin/pull/127

yahavi:
@kutran @a1flecke @OwenMoriarty-JNJ,
JFrog IDEA plugin 1.9.0 is released. This version includes support for Android Studio and Kotlin DSL.
Feel free to download and experience with it.
We'd appreciate your feedback on that.

PR #59, opened 2020-04-18
Support WebStorm and GoLand

yahavi:
Remove the obstacles to support _WebStorm_ and _GoLand_.

* Remove dependency on Java plugins by making 'maven', 'gradle', 'java' and 'groovy' plugins optional. This also will turn off inspections respectively.
* Catch `NoClassDefFoundError` when checking if Maven and Gradle are applicable. This is our sign that 'maven' or 'gradle' plugins are not installed.
* Fix the "unsupported" message.

With all of the above the plugin can work on _WebStorm_ and _GoLand_. During the next release we will be able to mark _WebStorm_ and _GoLand_ as supported in the marketplace.

**_Android Studio_**
Sadly, _Android Studio_ does not build Gradle dependencies tree as Intellij Idea does. However, this PR may take us one step forward toward supporting it in the future, as requested in #55.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -36,7 +36,7 @@ repositories {
 }
 
 dependencies {
-    implementation group: 'com.jfrog.ide', name: 'ide-plugins-common', version: '1.2.x-SNAPSHOT'
+    implementation group: 'com.jfrog.ide', name: 'ide-plugins-common', version: '1.3.0'
     compile group: 'org.apache.commons', name: 'commons-lang3', version: '3.9'
 }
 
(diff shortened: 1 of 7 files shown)


dlsniper:
Hello, please let me know if you need anything from the GoLand team at JetBrains to help with this and I'll be happy to bridge it. I'd love to see this released for our users.

eyalbe4:
Thank you so much @dlsniper!
I think @yahavi would appreciate some code review if anyone is available to do it (@yahavi, feel free to correct me or comment).
We're trying to reuse the the code of the existing JFrog Plugin for IntelliJ, to also support Goland (and WebStorm). We'd happy to get some confirmation from you guys that we're doing the right way. Any other suggestions, functional or other, would be greatly appreciated.

PR #66, opened 2020-06-03
Add functionality to exclude Maven transitive dependencies from issues-tree

barbelity:
Add the ability to exclude project's transitive dependencies from the Issues-Tree.
This first step applies only for Maven projects.

**Selecting the dependency to exclude:**
image

**Issues-tree is refreshed after exclusion:**
image

**pom.xml with exclusion:**
image

### DIFF
--- a/src/main/java/com/jfrog/ide/idea/exclusion/MavenExclusion.java
+++ b/src/main/java/com/jfrog/ide/idea/exclusion/MavenExclusion.java
@@ -1,0 +1,87 @@
+package com.jfrog.ide.idea.exclusion;
+
+import com.intellij.openapi.command.WriteCommandAction;
+import com.intellij.openapi.project.Project;
+import com.intellij.psi.PsiFile;
+import com.intellij.psi.PsiManager;
+import com.intellij.psi.xml.XmlFile;
+import com.intellij.psi.xml.XmlTag;
+import com.jfrog.ide.idea.inspections.MavenInspection;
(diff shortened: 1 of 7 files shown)


src/main/java/com/jfrog/ide/idea/exclusion/MavenExclusion.java
```diff
@@ -0,0 +1,87 @@
...
+        for (XmlTag exclusionTag : allExclusions) {
+            XmlTag groupIdTag = exclusionTag.findFirstSubTag(MavenInspection.MAVEN_GROUP_ID_TAG);
+            if (groupIdTag == null || !groupId.equals(groupIdTag.getValue().getText())) {
+                continue;
+            }
+            XmlTag artifactIdTag = exclusionTag.findFirstSubTag(MavenInspection.MAVEN_ARTIFACT_ID_TAG);
+            if (artifactIdTag == null || !artifactId.equals(artifactIdTag.getValue().getText())) {
+                continue;
+            }
+            return true;
```
yahavi, line 74:
```suggestion
            if (artifactIdTag != null && artifactId.equals(artifactIdTag.getValue().getText())) {
                return true;
            }
```

barbelity: Taken 👍

barbelity:
@yahavi I committed a fix for your review comments and the bug you found.
In addition, added a 'jump' to the corresponding dependency which the exclusion affects.
Tests will be added in a new PR.

PR #68, opened 2020-06-11
Add tests for Maven exclusion

barbelity:


barbelity:
> I also saw some IDE warnings on MavenExclusion:
> image
> image
> 
> Can you clean them also?

Thanks for that, my version of IntelliJ doesn't show these warnings 👍

PR #69, opened 2020-06-14
Fix exclusion of Maven dependencies + remove duplicated navigation targets for Gradle packages

barbelity:


src/main/java/com/jfrog/ide/idea/navigation/NavigationService.java
```diff
@@ -48,15 +48,32 @@ public void addNavigation(DependenciesTree treeNode, PsiElement navigationTarget
...
+     * @return true if the new navigation candidate is redundant.
+     */
+    private boolean isRedundantNavigation(Set<NavigationTarget> existingNavigations, NavigationTarget newNavigation) {
+        for (NavigationTarget existingTarget : existingNavigations) {
+            if (existingTarget.getLineNumber() == newNavigation.getLineNumber() &&
+                    existingTarget.getElement().getContainingFile().equals(newNavigation.getElement().getContainingFile())) {
+                return true;
+            }
+        }
+        return false;
```
yahavi, line 76: Instead of doing all of this, can we just change the `equals` and `hashCode` of `NavigationTarget` to compare the containing file of the PSI? I mean:
```java
    @Override
    public boolean equals(Object o) {
        if (this == o) return true;
        if (!(o instanceof NavigationTarget)) return false;
        NavigationTarget that = (NavigationTarget) o;
        return lineNumber == that.lineNumber &&
                Objects.equals(element.getContainingFile(), that.element.getContainingFile());
    }

    @Override
    public int hashCode() {
        return Objects.hash(element.getContainingFile(), lineNumber);
    }
```

barbelity: @yahavi in the beginning that's exactly what I did, but then I thought that it might be confusing in other (future) usages, as such behaviour isn't what you usually expect.
WDYT?

yahavi: I think the main reason to use set is to avoid duplication. The second reason is to insert and remove in O(1) using hash set.
For me, if you are using set you want to satisfy the both requirements above.
We miss both when we are using set with duplications and using a `for` loop to iterate in O(n).

barbelity: Ok, let's do this.

PR #70, opened 2020-06-14
Remove usages of deprecated APIs

barbelity:


barbelity:
@yahavi Implemented your suggestions 👍

PR #71, opened 2020-06-15
Configuration page - restore previously set credentials when selection changes on environment-variables checkbox

barbelity:
- [v] All [tests](https://ci.appveyor.com/project/jfrog-ecosystem/jfrog-idea-plugin) passed. If this feature is not already covered by the tests, I added new tests.
-----


src/main/java/com/jfrog/ide/idea/ui/configuration/XrayGlobalConfiguration.java
```diff
@@ -84,6 +84,10 @@ public XrayGlobalConfiguration() {
                 username.setEnabled(true);
                 url.setEnabled(true);
                 password.setEnabled(true);
+                XrayServerConfigImpl oldConfig = GlobalSettings.getInstance().getXrayConfig();
```
eyalbe4, line 87: Let's add a comment describing why we do this.

barbelity: Added 👍

PR #89, opened 2020-07-26
Configuration UI form improvements

### LINKED ISSUE #85: General usability suggestions
Hey guys :)
Few small yet important things I stumbled upon when using the plugin:
1. When no Xray server configured, the line `To Start using the JFrog Plugin, pleaseconfigure your JFrog Xray details` is missing a comma between `please` and `configure` link
2. When setting up the Xray URL, the first idea that comes to mind if to set the domain, i.e http://myartifactory.ac.me
However, that would result in `HTTP 404`.
I think an example URL should be added below the URL text input, saying `add /xray as a suffix to your Artifactory platform URL`
3. When setting a wrong URL, the error is cut, and the text is not fully shown (it should be wrapped inside the window itself)

yahavi:
-----
Resolves #85
1. Change the default background URL text of the Xray URL from: `Example: http://localhost:8000` to `http://myjfrog.acme.org/xray`:
Before: 
image
After: 
image

2. Wrap connection results text after clicking on the test connection button.
Before: 
image
After: 
image

### DIFF
--- a/src/main/java/com/jfrog/ide/idea/ui/configuration/XrayGlobalConfiguration.form
+++ b/src/main/java/com/jfrog/ide/idea/ui/configuration/XrayGlobalConfiguration.form
@@ -83,9 +83,7 @@
       </component>
       <component id="8234b" class="javax.swing.JLabel" binding="connectionResults">
         <constraints>
-          <grid row="6" column="1" row-span="1" col-span="2" vsize-policy="0" hsize-policy="0" anchor="8" fill="0" indent="0" use-parent-layout="false">
-            <maximum-size width="550" height="-1"/>
-          </grid>
+          <grid row="6" column="1" row-span="1" col-span="2" vsize-policy="0" hsize-policy="3" anchor="8" fill="0" indent="0" use-parent-layout="false"/>
         </constraints>
         <properties>
           <enabled value="true"/>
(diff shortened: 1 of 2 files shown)


src/main/java/com/jfrog/ide/idea/ui/configuration/XrayGlobalConfiguration.form
```diff
@@ -83,7 +83,9 @@
       </component>
       <component id="8234b" class="javax.swing.JLabel" binding="connectionResults">
         <constraints>
-          <grid row="6" column="1" row-span="1" col-span="2" vsize-policy="0" hsize-policy="0" anchor="8" fill="0" indent="0" use-parent-layout="false"/>
+          <grid row="6" column="1" row-span="1" col-span="2" vsize-policy="0" hsize-policy="0" anchor="8" fill="0" indent="0" use-parent-layout="false">
```
barbelity, line 86:
```suggestion
          <grid row="6" column="1" row-span="1" col-span="2" vsize-policy="0" hsize-policy="3" anchor="8" fill="0" indent="0" use-parent-layout="false">
```
I noticed that when this cell resizes due to a long message, it changes the whole window's size the right-hand part of the configurations window gets trimmed.
I suggest changing the 'hsize-policy' to better deal with long messages.

yahavi: Right!

PR #107, opened 2021-02-07
Maven - allow duplicated dependencies in same modules

### LINKED ISSUE #106: Maven dependancy difference
When scanning the exact same Maven project, different results/vulnerabilities are shown in the JFrog-supported IDEs. It appears that Eclipse has the most accurate data and number of dependancies, and that IntelliJ (and VSCode) provide different results, likely due to certain dependancies not showing up.

As the JFrog IntelliJ plugin is limited to what the IDE (IntelliJ) allows it to view/use, the purpose of this ticket is to investigate if anything can be done (on the plugin level) to improve this behavior and achieve uniformity across all JFrog-supported IDEs.

**Screenshot**
comparison

**Versions**
- JFrog IDEA plugin version: 1.7.0
- Operating system: Windows Server 2019
- Xray version: 3.x

yahavi:
-----
Partially resolved #106 

Currently duplicated Maven dependencies in the same module are not shows in the dependency tree.
This PR disallows this only 2 dependencies are descendants (to avoid loops).
Before:
Screen Shot 2021-02-07 at 11 46 38

After:
Screen Shot 2021-02-07 at 12 08 13

### DIFF
--- a/src/main/java/com/jfrog/ide/idea/scan/MavenScanManager.java
+++ b/src/main/java/com/jfrog/ide/idea/scan/MavenScanManager.java
@@ -148,6 +148,7 @@ public class MavenScanManager extends ScanManager {
     }
 
     private void updateChildrenNodes(DependenciesTree parentNode, MavenArtifactNode mavenArtifactNode, Set<String> added, boolean setScopes) {
+        // This set is used to disallow duplications between a node and its ancestors
         final Set<String> addedInSubTree = Sets.newHashSet(added);
         MavenArtifact mavenArtifact = mavenArtifactNode.getArtifact();
         DependenciesTree currentNode = new DependenciesTree(mavenArtifact.getDisplayStringSimple());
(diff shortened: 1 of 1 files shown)


src/main/java/com/jfrog/ide/idea/scan/MavenScanManager.java
```diff
@@ -148,6 +148,7 @@ private DependenciesTree populateMavenModuleNode(MavenProject mavenProject) {
     }
 
     private void updateChildrenNodes(DependenciesTree parentNode, MavenArtifactNode mavenArtifactNode, Set<String> added, boolean setScopes) {
+        final Set<String> addedInSubTree = Sets.newHashSet(added);
```
eyalbe4, line 151: Let's add a description for this set, explaining its function.

eyalbe4 (changes requested):

PR #127, opened 2021-05-10
Support Android Studio and Kotlin DSL

yahavi:
-----

* Add support for Android Studio by scan Gradle dependencies using a Gradle init script.
* Support Kotlin DSL in Gradle inspections.
* Support scan exclusions on Gradle projects.


src/main/java/com/jfrog/ide/idea/scan/ScanManagersFactory.java
```diff
@@ -121,7 +101,6 @@ public void refreshScanManagers() throws IOException {
             scanManagers.put(projectHash, scanManager);
         } else {
             createScanManagerIfApplicable(scanManagers, projectHash, ScanManagerTypes.MAVEN, "");
```
eyalbe4, line 103: Let's add a description here, explaining why the maven scan manager is created differently.

eyalbe4 (changes requested):

PR #169, opened 2021-10-31
Improve global configuration page

yahavi:
-----

* Split to tabs
* Add explanation for projects

Connection Details tab:
image

Advanced tab:
image


eyalbe4 (changes requested):
Some UI comments:
1.  "Project nane" should be replaced with ""Project key" both in the field label and the field description. Notice that in the description, the fields is referenced as "Project" and it should be changed to "Project key".
2. Let's replace the "JFrog Project" title with "Scan Policies".

I didn't go over the code yet. I'll let you know when done.

eyalbe4 (approved):

PR #176, opened 2021-11-16
Add access token support

talarian1:
-----

### DIFF
--- a/src/main/java/com/jfrog/ide/idea/configuration/GlobalSettings.java
+++ b/src/main/java/com/jfrog/ide/idea/configuration/GlobalSettings.java
@@ -32,8 +32,7 @@ import org.jetbrains.annotations.NotNull;
 import java.io.IOException;
 
 import static com.jfrog.ide.idea.ui.configuration.Utils.migrateXrayConfigToPlatformConfig;
-import static org.apache.commons.lang3.StringUtils.isAnyBlank;
-import static org.apache.commons.lang3.StringUtils.isBlank;
+import static org.apache.commons.lang3.StringUtils.*;
 
 /**
  * @author yahavi
@@ -173,6 +172,7 @@ public final class GlobalSettings implements PersistentStateComponent<GlobalSett
         setCommonConfigFields(serverConfig);
         this.serverConfig.setUsername(serverConfig.getUsername());
         this.serverConfig.setPassword(serverConfig.getPassword());
+        this.serverConfig.setAccessToken(serverConfig.getAccessToken());
         this.serverConfig.addCredentialsToPasswordSafe();
     }
 
@@ -207,6 +207,7 @@ public final class GlobalSettings implements PersistentStateComponent<GlobalSett
     private void migrateCredentialsFromFileToPasswordSafe(ServerConfigImpl serverConfig) {
         this.serverConfig.setUsername(serverConfig.getUsername());
         this.serverConfig.setPassword(serverConfig.getPassword());
+        this.serverConfig.setAccessToken(serverConfig.getAccessToken());
         this.serverConfig.addCredentialsToPasswordSafe();
         Application application = ApplicationManager.getApplication();
         application.invokeLater(application::saveSettings);
@@ -219,7 +220,7 @@ public final class GlobalSettings implements PersistentStateComponent<GlobalSett
      * @return true if credentials are stored in file.
      */
     private boolean shouldPerformCredentialsMigration(ServerConfigImpl xrayConfig) {
-        return !isAnyBlank(xrayConfig.getUsername(), xrayConfig.getPassword());
+        return isNoneBlank(xrayConfig.getUsername(), xrayConfig.getPassword());
     }
 
     /**
@@ -233,7 +234,8 @@ public final class GlobalSettings implements PersistentStateComponent<GlobalSett
             return;
         }
         try {
-            serverConfig.readConnectionDetailsFromJfrogCli();
+            String configured = serverConfig.readConnectionDetailsFromJfrogCli() ? "Successfully" : "Couldn't";
+            Logger.getInstance().info(configured + " config connection details from JFrog CLI");
         } catch (IOException exception) {
             Logger.getInstance().debug("Couldn't config connection details from JFrog CLI: " + ExceptionUtils.getRootCauseMessage(exception));
         }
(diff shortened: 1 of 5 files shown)


src/main/java/com/jfrog/ide/idea/configuration/GlobalSettings.java
```diff
@@ -203,18 +207,32 @@ public boolean areArtifactoryCredentialsSet() {
...
+        return isNoneBlank(xrayConfig.getUsername(), xrayConfig.getPassword());
+    }
+
+    public void loadConnectionDetailsFromJfrogCli() {
+        // Try to read connection details using JFrog CLI only if no server is already configured.
+        if (areXrayCredentialsSet()) {
+            return;
+        }
+        try {
+            String configured = serverConfig.readConnectionDetailsFromJfrogCli() ? "Successfully" : "Couldn't";
```
eyalbe4, line 232: Doesn't ```serverConfig.readConnectionDetailsFromJfrogCli``` return a boolean?

src/main/java/com/jfrog/ide/idea/configuration/GlobalSettings.java
```diff
@@ -203,18 +207,32 @@ public boolean areArtifactoryCredentialsSet() {
...
+            Logger.getInstance().info(configured + " config connection details from JFrog CLI");
```
eyalbe4, line 233: The format of this log message seem strange, becuase itstarts with the ```configured``` string.

src/main/java/com/jfrog/ide/idea/configuration/ServerConfigImpl.java
```diff
@@ -61,8 +64,10 @@
     static final String XRAY_URL_ENV = "JFROG_IDE_XRAY_URL";
     static final String USERNAME_ENV = "JFROG_IDE_USERNAME";
     static final String PASSWORD_ENV = "JFROG_IDE_PASSWORD";
+    static final String ACCESS_TOKEN_ENV = "JFROG_IDE_ACCESS_TOKEN";
     static final String PROJECT_ENV = "JFROG_IDE_PROJECT";
 
+
```
eyalbe4, line 70: Redundant new-line.

src/main/java/com/jfrog/ide/idea/configuration/ServerConfigImpl.java
```diff
@@ -117,11 +125,15 @@
...
-        return !isAnyBlank(xrayUrl, username, password);
+        return isNotBlank(xrayUrl) && isAuthenticationConfigured();
     }
 
     boolean isArtifactoryConfigured() {
-        return !isAnyBlank(artifactoryUrl, username, password);
+        return isNotBlank(artifactoryUrl) && isAuthenticationConfigured();
+    }
+
+    boolean isAuthenticationConfigured() {
```
eyalbe4, line 135: Looks like this method can be made private.

src/main/java/com/jfrog/ide/idea/configuration/ServerConfigImpl.java
```diff
@@ -346,12 +365,21 @@ public void setPassword(String password) {
         this.password = password;
     }
 
+    public void setAccessToken(String accessToken) {
+        this.accessToken = accessToken;
+    }
+
+
```
eyalbe4, line 372: Redundant new-line.

src/main/java/com/jfrog/ide/idea/ui/configuration/JFrogGlobalConfiguration.java
```diff
@@ -336,8 +385,15 @@ private void updateConnectionDetailsTextFields() {
...
-        username.setText(serverConfig.getUsername());
-        password.setText(serverConfig.getPassword());
+        if (isNotBlank(serverConfig.getAccessToken())) {
+            accessToken.setText(serverConfig.getAccessToken());
+            accessTokenRadioButton.setSelected(true);
+
+        } else {
+            username.setText(serverConfig.getUsername());
+            password.setText(serverConfig.getPassword());
+            //usernamePasswordRadioButton.setSelected(true);
```
eyalbe4, line 395: Let's avoid keeping commented out code.

src/main/java/com/jfrog/ide/idea/ui/configuration/JFrogGlobalConfiguration.java
```diff
@@ -336,8 +385,15 @@ private void updateConnectionDetailsTextFields() {
         platformUrl.setText(serverConfig.getUrl());
         xrayUrl.setText(serverConfig.getXrayUrl());
         artifactoryUrl.setText(serverConfig.getArtifactoryUrl());
-        username.setText(serverConfig.getUsername());
-        password.setText(serverConfig.getPassword());
+        if (isNotBlank(serverConfig.getAccessToken())) {
+            accessToken.setText(serverConfig.getAccessToken());
+            accessTokenRadioButton.setSelected(true);
+
```
eyalbe4, line 391: redundant new-line.

eyalbe4 (changes requested):

src/main/java/com/jfrog/ide/idea/scan/ScanManager.java
```diff
@@ -150,10 +150,6 @@ public void run(@NotNull com.intellij.openapi.progress.ProgressIndicator indicat
                 if (project.isDisposed()) {
                     return;
                 }
-                if (!GlobalSettings.getInstance().areXrayCredentialsSet()) {
-                    getLog().warn("Xray server is not configured.");
-                    return;
-                }
```
yahavi, line 156: What is the reason for removing this?

talarian1: Because as much as I saw it was redundant.
https://github.com/jfrog/jfrog-idea-plugin/blob/master/src/main/java/com/jfrog/ide/idea/scan/ScanManagersFactory.java#L87-L90
What do you think?

src/main/java/com/jfrog/ide/idea/configuration/GlobalSettings.java
talarian1: This way it works. Do you prefer a standard if and duplicated log message print?

src/main/java/com/jfrog/ide/idea/scan/ScanManager.java
yahavi: Scans can be triggered also by a change in `package-json.lock`, `go.mod`, `pom.xml`, and more. Scans triggered by a change in the project descriptor do not need the scan managers' factory.
We would prefer to avoid triggering scans when the credentials are not configured, since it performs heavy operations (like building the dependency tree) for nothing.

talarian1: I'm not sure it is a valid scenario after we added the new JFrog CLI config logic. Anyway, are you fine with the original double-check?

PR #185, opened 2021-12-21
Show severity icons near dependencies

yahavi:
-----

* Instead of showing red or white underlines beneath vulnerable dependencies, show the severity icon near them.
image

* Get rid of the meaningless records in the problems views:
image


src/main/java/com/jfrog/ide/idea/inspections/AbstractInspection.java
```diff
@@ -117,6 +117,16 @@ void visitElement(AnnotationHolder annotationHolder, PsiElement element) {
      */
     abstract Set<DependencyTree> getModules(PsiElement element, String componentName);
 
+    /**
+     * Override this method to avoid displaying multiple annotation icons in the same line.
```
eyalbe4, line 121: ... to avoid displaying... --> ... to determine whether to display...

eyalbe4 (changes requested):

PR #189, opened 2022-01-23
Support Xray watches

yahavi:
-----
Depends on https://github.com/jfrog/ide-plugins-common/pull/68 & https://github.com/jfrog/xray-client-java/pull/30

image


src/main/java/com/jfrog/ide/idea/ui/configuration/JFrogGlobalConfiguration.java
```diff
@@ -399,4 +446,25 @@ private void updateConnectionDetailsTextFields() {
...
+                showAccordingToProjectRadioButton.setSelected(true);
+                watch.setEnabled(false);
+                return;
+            case VULNERABILITIES:
+                showAllPoliciesRadioButton.setSelected(true);
+                watch.setEnabled(false);
+        }
+    }
+
+    private void createUIComponents() {
```
talarian1, line 467: Is this one needed?

talarian1:
Looks great to me! :) 
However, maybe we should support a list of watches (like we do in the JFrog CLI)? I think it is a really small change that may be useful for some users.

PR #201, opened 2022-02-19
Root project should not be excluded from scanning

yahavi:
-----

Resolves https://github.com/jfrog/jfrog-idea-plugin/issues/198 & https://github.com/jfrog/jfrog-idea-plugin/issues/108

Adapt to changes introduced in https://github.com/jfrog/ide-plugins-common/pull/72


src/main/java/com/jfrog/ide/idea/scan/ScanManagersFactory.java
```diff
@@ -163,7 +162,9 @@ private ScanLogic createScanLogic(Utils.ScanLogicType type, ScanCache scanCache,
     }
 
     private void createScanManagers(Map<Integer, ScanManager> scanManagers, Set<Path> scanPaths, ExecutorService executor) throws IOException {
```
eyalbe4, line 164: I think we better add a description for this method. Specifically describe what the ```scanPaths``` set includes.

eyalbe4 (changes requested):

PR #205, opened 2022-03-13
Better handling UI events

yahavi:
-----

Fix the following UI issues:
* Fix an issue whereby reading credentials from CLI doesn't initiate the "More Info" panel.
* Fix an issue whereby starting a new scan doesn't reset the dependency tree and the issues table.

Update supported version from `IU-2020.3.1` to `IU-2020.3.2`, in order to handle an exception thrown in the latest macOS:
> java.lang.ClassNotFoundException: com.apple.eawt.OpenURIHandler


src/main/java/com/jfrog/ide/idea/scan/ScanManagersFactory.java
```diff
@@ -113,6 +107,21 @@ public void startScan(boolean quickScan) {
         }
     }
 
+    private boolean refreshConnection() {
```
eyalbe4, line 110: Let's add a description for method. Specifically explain the boolean returned value.

eyalbe4 (changes requested):

PR #246, opened 2022-12-19
Update release notes

yahavi:
-----


eyalbe4 (changes requested):
Let's use "Bug" instead of "Bugfix".

PR #247, opened 2022-12-20
Add Gradle task to download the new webview

yahavi:
-----

* Add 3 new Gradle tasks:
1. 'downloadWebview' - Download the Webview tar.gz file and trigger 'verifyWebview'
2. 'verifyWebview' - Verify sha512 of the downloaded tar.gz file and trigger 'extractWebview'
3. 'extractWebview' - Extract the tar.gz file to `src/main/resources/webview/`
* Make the 'compileJava' task depends on the 'downloadWebview' to make all of this added functionality transparent to the plugin developer


build.gradle
```diff
@@ -65,9 +68,35 @@ test {
...
+    src 'https://releases.jfrog.io/artifactory/ide-webview-npm/jfrog-ide-webview/-/' + webviewFileName
+    dest buildDir
+    onlyIfModified true
+    finalizedBy('verifyWebview')
+}
+
+tasks.register('verifyWebview', Verify) {
+    src new File(buildDir, webviewFileName)
+    algorithm 'SHA-512'
+    checksum 'b2015294714222de1efa2490e9d1b8d36fd82aa79a4d04fd0821c07da8799eb85d45f1dedede2366da71b2d91f6277e31ce7befc9c4cbe2229b437d731fc7689'
```
Or-Geva, line 83: If the checksum does not match, it returns an error. If the webview version has been updated or it is the first time running this task, an error occurs.

build.gradle
```diff
@@ -65,9 +68,35 @@ test {
...
+    finalizedBy('extractWebview')
+}
+
+tasks.register('extractWebview', Copy) {
+    from tarTree(new File(buildDir, webviewFileName))
+    into Paths.get('src', 'main', 'resources', 'jfrog-ide-webview').toFile()
+    eachFile {
```
Or-Geva, line 90: Only the build directory in the .tgz file needs to be copied

Or-Geva (changes requested):

build.gradle
yahavi: We (me and Or) agreed not to change this logic at this time, since we didn't found more secured solution

Issue #99, opened 2020-12-15
Request for PyCharm integration


eranbl:
This is a request to add IDE integration for PyCharm.

yahavi:
@eranbl,
PyCharm and PyPI projects are supported in the new releases, 1.9.0 and 1.9.1.

PR #261, opened 2023-01-03
Added ScannerSearchTarget info for the contextual analysis view

talarian1:
-----
Depends on https://github.com/jfrog/ide-plugins-common/pull/99

### DIFF
--- a/src/main/java/com/jfrog/ide/idea/scan/data/Rule.java
+++ b/src/main/java/com/jfrog/ide/idea/scan/data/Rule.java
@@ -45,7 +45,7 @@ public class Rule {
 
     @Override
     public int hashCode() {
-        return Objects.hash(id, shortDescription, fullDescription);
+        return Objects.hash(id);
     }
 
     @Override
@@ -57,6 +57,6 @@ public class Rule {
             return false;
         }
         Rule rhs = ((Rule) other);
-        return Objects.equals(this.id, rhs.id) && Objects.equals(this.shortDescription, rhs.shortDescription) && Objects.equals(this.fullDescription, rhs.fullDescription);
+        return Objects.equals(this.id, rhs.id);
     }
 }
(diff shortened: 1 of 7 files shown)


src/main/java/com/jfrog/ide/idea/scan/data/Rule.java
```diff
@@ -30,9 +33,19 @@ public void setShortDescription(Message shortDescription) {
...
+
+    @SuppressWarnings({"unused"})
+    public void setFullDescription(Message fullDescription) {
+        this.fullDescription = fullDescription;
+    }
+
     @Override
     public int hashCode() {
-        return Objects.hash(id, shortDescription);
+        return Objects.hash(id, shortDescription, fullDescription);
```
Or-Geva, line 48: Isn't the 'id' unique enough?

src/main/java/com/jfrog/ide/idea/ui/webview/WebviewObjectConverter.java
```diff
@@ -30,15 +30,14 @@ public static DependencyPage convertIssueToDepPage(IssueNode issueNode) {
...
-                    }
-                    applicableDetails = new ApplicableDetails(true, evidences, null);
+                String searchTarget = applicableIssues.get(0).getScannerSearchTarget();
+                Evidence[] evidences = new Evidence[applicableIssues.size()];
+                int i = 0;
+                for (ApplicableIssueNode applicableIssue : applicableIssues) {
+                    evidences[i++] = new Evidence(applicableIssue.getReason(), applicableIssue.getFilePath(), applicableIssue.getLineSnippet());
                 }
+                applicableDetails = new ApplicableDetails(true, evidences, searchTarget);
+
```
Or-Geva, line 40: This logic should be extracted into a dedicated function (or class?). Also, writing tests for this logic will be easier.

Or-Geva (changes requested):

PR #262, opened 2023-01-05
Fix file descriptors inspections

talarian1:
-----
Depends on https://github.com/jfrog/ide-plugins-common/pull/100


asafgabai (changes requested):

asafgabai (changes requested):

src/main/java/com/jfrog/ide/idea/inspections/AbstractInspection.java
```diff
@@ -168,207 +178,61 @@ boolean isShowInspection(PsiElement element) {
...
-    /**
-     * Get all submodules from the dependencies-tree which are containing the dependency element.
-     * Currently in use for Gradle and Maven.
+     * Get the dependencies that match to the input componentName.
      *
-     * @param root          - The root of the dependency tree
-     * @param project       - The project
-     * @param modulesList   - List of all relevant modules
+     * @param file          - The Descriptor file node in the tree
      * @param componentName - Component name representing a dependency without version
```
asafgabai, line 209: Why actually don't we compare also the component version? In most cases, we can get it from the PsiElement.

src/main/java/com/jfrog/ide/idea/inspections/AbstractInspection.java
```diff
@@ -168,207 +178,61 @@ boolean isShowInspection(PsiElement element) {
...
-     * @return true if the general info matches the component name
+     * @return true if the node matches the component name
      */
-    boolean compareGeneralInfo(GeneralInfo generalInfo, String componentName) {
-        // Module name
+    boolean isNodeMatch(DependencyNode node, String componentName) {
+        String artifactID = node.getGeneralInfo().getComponentIdWithoutPrefix();
+        ImpactTreeNode impactPath = node.getImpactPaths();
         if (StringUtils.countMatches(componentName, ":") == 0) {
```
asafgabai, line 232: In what scenarios it's false? I'd suggest adding a comment explaining it.

asafgabai (changes requested):

PR #263, opened 2023-01-09
V1: Skip loading cache from file-system in full scan mode

yahavi:
-----

Resolves https://github.com/jfrog/jfrog-idea-plugin/issues/248
Depends on https://github.com/jfrog/ide-plugins-common/pull/101

When changing watches, the cache does not get updated.
This PR changes the way we initialize the scan cache in case of a full scan, by avoiding reading the cache from the file system.

### DIFF
--- a/src/main/java/com/jfrog/ide/idea/scan/ScanManagersFactory.java
+++ b/src/main/java/com/jfrog/ide/idea/scan/ScanManagersFactory.java
@@ -94,7 +94,7 @@ public class ScanManagersFactory implements Disposable {
         project.getMessageBus().syncPublisher(ApplicationEvents.ON_SCAN_LOCAL_STARTED).update();
         ExecutorService executor = Executors.newFixedThreadPool(3);
         try {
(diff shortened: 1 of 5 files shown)


src/main/java/com/jfrog/ide/idea/scan/ScanManagersFactory.java
```diff
@@ -246,26 +246,29 @@ private void createScanManagersForPackageDirs(Set<String> packageDirs, Map<Integ
...
 
     /**
      * Create the scan cache object and the directories needed for it.
      *
+     * @param quickScan - True to load the scan cache.
      * @return scan cache.
      * @throws IOException in cace of any I/O error.
      */
-    private ScanCache createXrayScanCache() throws IOException {
+    private ScanCache createXrayScanCache(boolean quickScan) throws IOException {
```
talarian1, line 267: I think that here the `quickScan` is a bit confusing. What do you think about something like `loadFromDisk` or `initNew`?

talarian1 (approved):

PR #265, opened 2023-01-09
Support  usage report with access token

talarian1:
-----
Fixes usage report sending when authentication is done with an access token.


yahavi (approved):
Good catch!

PR #275, opened 2023-01-17
Reimplement jump from tree to editor

talarian1:
-----

### DIFF
--- a/src/main/java/com/jfrog/ide/idea/ui/ComponentsTree.java
+++ b/src/main/java/com/jfrog/ide/idea/ui/ComponentsTree.java
@@ -32,9 +32,7 @@ public abstract class ComponentsTree extends Tree {
     protected Project project;
     protected final List<ToolbarPopupMenu> toolbarPopupMenus = new ArrayList<>();
     protected final JBPopupMenu popupMenu = new JBPopupMenu();
-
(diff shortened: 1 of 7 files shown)


asafgabai (changes requested):

src/main/java/com/jfrog/ide/idea/ui/ComponentsTree.java
```diff
@@ -71,35 +73,24 @@ private void handleContextMenu(ComponentsTree tree, MouseEvent e) {
...
-        if (selPath == null) {
+        TreePath selectedPath = tree.getPathForLocation(e.getX(), e.getY());
+        if (selectedPath == null) {
             return;
         }
-        createNodePopupMenu((DependencyTree) selPath.getLastPathComponent());
-        popupMenu.show(tree, e.getX(), e.getY());
+        if (selectedPath.getLastPathComponent() instanceof DependencyNode) {
+            createNodePopupMenu((DependencyNode) selectedPath.getLastPathComponent());
+            popupMenu.show(tree, e.getX(), e.getY());
```
yahavi, line 83: For better user experience - let's show the menu when clicking not only on the component text, but also in the right of it

yahavi (changes requested):

PR #276, opened 2023-01-17
Automatically update the Webview checksum

yahavi:
-----

After updating the Webview version, Gradle will update the expected sha256 of the Webview when running locally.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -81,6 +81,7 @@ tasks.register('downloadWebview', Download) {
 tasks.register('getAndUpdateWebviewChecksum') {
     finalizedBy('verifyWebview')
     if (System.getenv("CI") != null) {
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -68,19 +69,29 @@ test {
...
+    src webviewUrl
     dest buildDir
     onlyIfModified true
+    finalizedBy('getAndUpdateWebviewChecksum')
+}
+
+tasks.register('getAndUpdateWebviewChecksum') {
     finalizedBy('verifyWebview')
+    if (System.getenv("CI") != null) {
+        ext.checksum = webviewChecksum
```
Or-Geva, line 84: Would it be possible to write log message here that the CI mode is active, or something similar?

PR #279, opened 2023-01-18
Upgrade the visualization of not applicable security issues

talarian1:
-----
Depends on https://github.com/jfrog/ide-plugins-common/pull/102

Support better visualization for not applicable security issues.
Adds severity icon notation to source code files as well.


asafgabai (changes requested):

src/main/java/com/jfrog/ide/idea/inspections/JFrogSecurityAnnotator.java
```diff
@@ -37,19 +39,31 @@ public PsiFile collectInformation(@NotNull PsiFile file, @NotNull Editor editor,
...
-                .orElse(null);
-        return scanManager != null ? scanManager.getSourceCodeScanResults() : new ArrayList<>();
+    public List<ApplicableIssueNode> doAnnotate(PsiFile file) {
+        List<ApplicableIssueNode> applicableIssues = new ArrayList<>();
+        ComponentsTree componentsTree = LocalComponentsTree.getInstance(file.getProject());
+        if (componentsTree == null || componentsTree.getModel() == null) {
+            return null;
+        }
+        Enumeration<TreeNode> roots = ((BaseTreeNode) componentsTree.getModel().getRoot()).children();
+        for (TreeNode root : Collections.list(roots)) {
```
yahavi, line 49: I'd prefer using the enumeration directly and not initializaing a new list. The new list duplicating the children of the tree and causes some redundant work.
Same for the for loop below.

src/main/java/com/jfrog/ide/idea/inspections/JFrogSecurityAnnotator.java
```diff
@@ -37,19 +39,31 @@ public PsiFile collectInformation(@NotNull PsiFile file, @NotNull Editor editor,
...
     }
 
     @Override
-    public void apply(@NotNull PsiFile file, List<JFrogSecurityWarning> warnings, @NotNull AnnotationHolder holder) {
-        String filePath = file.getVirtualFile() != null ? file.getVirtualFile().getPath() : null;
-        warnings.stream().filter(Objects::nonNull).filter(warning -> warning.isApplicable() && warning.getFilePath().equals(filePath)).forEach(warning -> {
-            int startOffset = StringUtil.lineColToOffset(file.getText(), warning.getLineStart(), warning.getColStart());
-            int endOffset = StringUtil.lineColToOffset(file.getText(), warning.getLineEnd(), warning.getColEnd());
+    public void apply(@NotNull PsiFile file, List<ApplicableIssueNode> warnings, @NotNull AnnotationHolder holder) {
+        warnings.stream().filter(Objects::nonNull).forEach(warning -> {
```
yahavi, line 62: Instead of the `if (document != null)` below
```suggestion
        Document document = PsiDocumentManager.getInstance(file.getProject()).getDocument(file);
        if (document == null) {
            return;
        }
        warnings.stream().filter(Objects::nonNull).forEach(warning -> {
```

yahavi (changes requested):

PR #280, opened 2023-01-22
Fix wrong Gradle home path

asafgabai:
-----
Fixed a bug that caused using wrong Gradle project settings. It happened in cases multiple Gradle projects are open in IntelliJ.


yahavi (approved):
LGTM

PR #283, opened 2023-01-31
Restructure

asafgabai:
-----
- Split ScanManagersFactory to ScannerFactory and ScanManager
- Class name changes (pay attention to ScanManager -> ScannerBase)
- Use one ScanLogic for all Scanners
- Source code scanning restructure


src/main/java/com/jfrog/ide/idea/scan/ScannerFactory.java
```diff
@@ -0,0 +1,192 @@
...
+
+    public static int getModuleIdentifier(String name, String path) {
+        return Objects.hashCode(name, path);
+    }
+
+    /**
+     * Scan projects, create new Scanners and delete unnecessary ones.
+     * Existing Scanners from previous scans, are not overridden.
+     */
+    public Map<Integer, ScannerBase> refreshScanners(Map<Integer, ScannerBase> oldScanners, ScanLogic scanLogic,
```
talarian1, line 43: I think we should consider renaming refreshScanners to create/initialize scanners.

src/main/java/com/jfrog/ide/idea/scan/ScannerFactory.java
```diff
@@ -0,0 +1,192 @@
...
+    /**
+     * Create npm, Gradle, Go and Yarn Scanners.
+     *
+     * @param newScanners the new Scanners map to add the Scanners into
+     * @param oldScanners the Scanners map including the Scanner of the current project, or an empty map for a fresh start
+     * @param scanPaths   potential paths for scanning for package descriptor files
+     * @param executor    the thread pool
+     * @throws IOException in case of any I/O error during the search for the actual package descriptor files.
+     */
+    private void refreshGenericScanners(Map<Integer, ScannerBase> newScanners, Map<Integer, ScannerBase> oldScanners,
```
talarian1, line 62: is there any advantage for the "output" param` newScanners`
(instead of returning it as a return value)?

talarian1 (changes requested):

src/main/java/com/jfrog/ide/idea/scan/ScannerFactory.java
asafgabai: But if a scanner already exists, it's not created again.

src/main/java/com/jfrog/ide/idea/scan/ScannerFactory.java
asafgabai: It's not exactly an output param. `newScanners` is created outside of this method, and even might contain Scanners in it.

src/main/java/com/jfrog/ide/idea/scan/ScannerBase.java
```diff
@@ -0,0 +1,376 @@
...
+            // Force inspections run due to changes in the displayed tree
+            runInspections();
+
+        } catch (ProcessCanceledException e) {
+            log.info("Xray scan was canceled");
+        } catch (Exception e) {
+            logError(log, "Xray Scan failed", e, true);
+        } finally {
+            scanInProgress.set(false);
+            sendUsageReport();
```
yahavi, line 174: I know it is not your change, but I think that the usage report should be sent asynchronously at the beginning of the scan.

src/main/java/com/jfrog/ide/idea/ui/webview/WebviewObjectConverter.java
```diff
@@ -206,4 +205,24 @@ private static Reference[] convertReferences(List<String> xrayReferences) {
             return new Reference(parts[1].substring(0, parts[1].length() - 1), parts[0].substring(1));
         }).toArray(Reference[]::new);
     }
+
+    private static String getPackageTypeName(String componentId) {
+        String GENERIC_PKG_TYPE = "Generic";
+        String[] compIdParts = componentId.split("://");
```
yahavi, line 211: You can use `StringUtils.substringBefore(componentId, "://");` to get the component prefix.

src/main/java/com/jfrog/ide/idea/ui/webview/WebviewObjectConverter.java
```diff
@@ -206,4 +205,24 @@ private static Reference[] convertReferences(List<String> xrayReferences) {
...
+            case "gav":
+                return "Maven";
+            case "go":
+                return "Go";
+            case "npm":
+                return "npm";
+            case "pypi":
+                return "PyPI";
+            default:
+                return GENERIC_PKG_TYPE;
```
yahavi, line 225: Let's export all return values to class consts.

src/main/java/com/jfrog/ide/idea/ui/webview/WebviewObjectConverter.java
```diff
@@ -206,4 +205,24 @@ private static Reference[] convertReferences(List<String> xrayReferences) {
...
         }).toArray(Reference[]::new);
     }
+
+    private static String getPackageTypeName(String componentId) {
+        String GENERIC_PKG_TYPE = "Generic";
+        String[] compIdParts = componentId.split("://");
+        if (compIdParts.length != 2) {
+            return GENERIC_PKG_TYPE;
+        }
+        switch (compIdParts[0]) {
```
yahavi, line 215: You can switch here, somehow on ComponentPrefix.valueOf(...) (requires some changes)

yahavi (changes requested):

src/main/java/com/jfrog/ide/idea/scan/ScanUtils.java
```diff
@@ -34,15 +34,15 @@ public class ScanUtils {
     public static boolean isLocalProjectSupported(Project project) throws IOException {
```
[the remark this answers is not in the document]
asafgabai: To avoid scanning the filesystem twice, I removed this method. This check will be added with the empty state handling soon.

src/main/java/com/jfrog/ide/idea/ui/webview/WebviewObjectConverter.java
asafgabai: But I won't know if the "://" separator exists. There are cases when it doesn't, and Xray refers to the project as Generic.

yahavi: In this case the `substringBefore` will return an empty string and you can return "Generic"

src/main/java/com/jfrog/ide/idea/scan/SourceCodeScannerManager.java
```diff
@@ -96,7 +96,7 @@ private List<String> getSkippedFoldersPatterns() {
         return skippedFoldersPatterns;
     }
 
-    public List<FileTreeNode> getResults(Map<String, List<IssueNode>> issuesMap) {
+    private List<FileTreeNode> groupResultsToFileTreeNodes(List<JFrogSecurityWarning> scanResults, Map<String, List<VulnerabilityNode>> issuesMap) {
```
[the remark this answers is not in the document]
asafgabai: I added javadoc and comments.

src/main/java/com/jfrog/ide/idea/scan/ScannerBase.java
```diff
@@ -136,6 +136,9 @@ protected void sendUsageReport() {
      */
     private void scanAndUpdate(ProgressIndicator indicator) {
         try {
+            // Send usage report
```
Or-Geva, line 139: Seems redundant.
It would be best if we avoid polluting the code.

src/main/java/com/jfrog/ide/idea/scan/SourceCodeScannerManager.java
```diff
@@ -96,14 +96,26 @@ private List<String> getSkippedFoldersPatterns() {
         return skippedFoldersPatterns;
     }
 
+    /**
+     * Convert {@link JFrogSecurityWarning}s to a list of {@link FileTreeNode}s with source code issues.
+     * Also, update {@link VulnerabilityNode}s with their matching source code issues.
```
Or-Geva, line 101: Since the name of the function, `groupResultsToFileTreeNodes` has nothing to do with updating, I recommend moving the update logic to another function.

In addition, the comment begins with the verb 'convert..', but the function name implies grouping. Anything that forces you to check the function body should be avoided.

src/main/java/com/jfrog/ide/idea/scan/SourceCodeScannerManager.java
```diff
@@ -96,14 +96,26 @@ private List<String> getSkippedFoldersPatterns() {
...
+     *
+     * @param scanResults a list of source code scan results.
+     * @param issuesMap   a map of {@link VulnerabilityNode}s mapped by their CVEs.
+     * @return a list of new {@link FileTreeNode}s containing source code issues.
+     */
     private List<FileTreeNode> groupResultsToFileTreeNodes(List<JFrogSecurityWarning> scanResults, Map<String, List<VulnerabilityNode>> issuesMap) {
+        // Map of new FileTreeNodes mapped by their paths
```
Or-Geva, line 108: It is best practice to replace comments with a meaningful variable name.

src/main/java/com/jfrog/ide/idea/scan/SourceCodeScannerManager.java
asafgabai: Moving the update logic to another method will make the code less efficient.
I made some changes to make this method more readable, changed its name, and fixed its javadoc.

PR #285, opened 2023-02-06
Add Inactive Jfrog platform handling (free-tier)

talarian1:
-----
Detect and suggest a solution for an inactive JFrog platform (free-tier).

Screenshot 2023-02-06 at 11 25 03

Fixes https://github.com/jfrog/jfrog-idea-plugin/issues/228
Depends on https://github.com/jfrog/xray-client-java/pull/42, https://github.com/jfrog/build-info/pull/698

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -52,7 +52,7 @@ dependencies {
     implementation group: 'com.jfrog.xray.client', name: 'xray-client-java', version: '0.13.x-20230219.134632-1'
     implementation group: 'org.apache.commons', name: 'commons-collections4', version: '4.4'
     implementation group: 'org.jfrog.filespecs', name: 'file-specs-java', version: '1.1.2'
-    implementation group: 'com.jfrog.ide', name: 'ide-plugins-common', version: '1.13.x-20230214.072934-1'
+    implementation group: 'com.jfrog.ide', name: 'ide-plugins-common', version: '1.13.x-20230221.095443-1'
     implementation group: 'org.apache.commons', name: 'commons-lang3', version: '3.11'
     implementation group: 'com.google.guava', name: 'guava', version: '30.1.1-jre'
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/jfrog/ide/idea/scan/ScanManagersFactory.java
```diff
@@ -84,13 +93,36 @@ public void startScan() {
...
+        } catch (JFrogInactiveEnvironmentException e) {
+            handleJfrogInactiveEnvironment(e.getRedirectUrl());
         } catch (IOException | RuntimeException | InterruptedException e) {
             logError(Logger.getInstance(), "", e, true);
         } finally {
             executor.shutdown();
         }
     }
 
+    private void handleJfrogInactiveEnvironment(String redirectUrl) {
```
yahavi, line 105: Let's add a JavaDoc above this method explaining what it does

src/main/java/com/jfrog/ide/idea/scan/ScanManagersFactory.java
```diff
@@ -84,13 +93,36 @@ public void startScan() {
...
+                        })
+                .setCloseButtonEnabled(true)
+                .setHideOnAction(true)
+                .setHideOnClickOutside(true)
+                .setHideOnLinkClick(true)
+                .setHideOnKeyOutside(true)
+                .setDialogMode(true)
+                .createBalloon();
+        balloon.show(RelativePoint.getNorthWestOf(statusBar.getComponent()), Balloon.Position.atRight);
+
```
yahavi, line 123: Redundant new line
```suggestion
```

src/main/java/com/jfrog/ide/idea/scan/ScanManagersFactory.java
```diff
@@ -84,13 +93,36 @@ public void startScan() {
                     logError(Logger.getInstance(), "", e, true);
                 }
             }
+        } catch (JFrogInactiveEnvironmentException e) {
+            handleJfrogInactiveEnvironment(e.getRedirectUrl());
```
yahavi, line 97: Shouldn't we do this also in the test connection button?
Same question for CI mode.

yahavi (changes requested):

src/main/java/com/jfrog/ide/idea/scan/ScanManagersFactory.java
talarian1: Added this to the test connection:
Screenshot 2023-02-19 at 18 15 56
Regarding the CI view, I don't think it is necessary - an Error will be displayed in the logs.

PR #286, opened 2023-02-09
Apply Connections details change between source code scanners runs

talarian1:
-----
The source code scanner should receive the latest connection details before each run.

### DIFF
--- a/src/main/java/com/jfrog/ide/idea/scan/ScanBinaryExecutor.java
+++ b/src/main/java/com/jfrog/ide/idea/scan/ScanBinaryExecutor.java
@@ -29,13 +29,11 @@ import static com.jfrog.ide.idea.utils.Utils.HOME_PATH;
  * @author Tal Arian
  */
 public abstract class ScanBinaryExecutor {
-    private static final Path BINARIES_DIR = HOME_PATH.resolve("dependencies").resolve("jfrog-security");
-    private CommandExecutor commandExecutor;
-    private final String BINARY_PATH;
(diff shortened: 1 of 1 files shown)


src/main/java/com/jfrog/ide/idea/scan/ScanBinaryExecutor.java
```diff
@@ -30,7 +30,8 @@
  */
 public abstract class ScanBinaryExecutor {
     private static final Path BINARIES_DIR = HOME_PATH.resolve("dependencies").resolve("jfrog-security");
-    private final CommandExecutor commandExecutor;
+    private CommandExecutor commandExecutor;
+    private final String BINARY_PATH;
```
yahavi, line 34: Let's use camel case, to be consistent with our conventions

src/main/java/com/jfrog/ide/idea/scan/ScanBinaryExecutor.java
```diff
@@ -30,7 +30,8 @@
  */
 public abstract class ScanBinaryExecutor {
     private static final Path BINARIES_DIR = HOME_PATH.resolve("dependencies").resolve("jfrog-security");
-    private final CommandExecutor commandExecutor;
+    private CommandExecutor commandExecutor;
```
yahavi, line 33: Can be converted to a local variable

yahavi (changes requested):

yahavi (approved):

PR #293, opened 2023-02-16
Cache last scan results

asafgabai:
-----
Save last scan results in cache and show them on plugin startup.


src/main/java/com/jfrog/ide/idea/ui/LocalComponentsTree.java
```diff
@@ -163,4 +170,25 @@ public void actionPerformed(ActionEvent e) {
             }
         });
     }
+
+    public void cacheTree() throws IOException {
+        SortableChildrenTreeNode root = (SortableChildrenTreeNode) getModel().getRoot();
+        cache.cacheNodes(Collections.list(root.children()).stream().map(treeNode -> (FileTreeNode) treeNode).collect(Collectors.toList()));
```
yahavi, line 176: More efficient - removes 2 passes on the root children:
```suggestion
        //noinspection unchecked
        cache.cacheNodes((List<FileTreeNode>) root.children());
```

src/main/java/com/jfrog/ide/idea/scan/ScanManager.java
```diff
@@ -65,23 +68,31 @@ public void startScan() {
...
-        try {
-            ScanLogic scanLogic = createScanLogic();
-            refreshScanners(scanLogic, executor);
-            NavigationService.clearNavigationMap(project);
-            for (ScannerBase scanner : scanners.values()) {
-                try {
-                    scanner.asyncScanAndUpdateResults();
-                } catch (RuntimeException e) {
-                    logError(Logger.getInstance(), "", e, true);
+        Thread currScanThread = new Thread(() -> {
```
yahavi, line 71: Disclaimer - Assaf and I agreed to leave it Thread, but may change it to a Backgroundable task after the release

src/main/java/com/jfrog/ide/idea/scan/ScanManager.java
```diff
@@ -23,11 +24,13 @@
 import java.util.Set;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 
 import static com.jfrog.ide.common.log.Utils.logError;
 import static com.jfrog.ide.common.utils.XrayConnectionUtils.createXrayClientBuilder;
 
 public class ScanManager {
+    private final long SCAN_TIMEOUT_MINUTES = 10;
```
yahavi, line 33: int requires less memory
```suggestion
    private final int SCAN_TIMEOUT_MINUTES = 10;
```

yahavi (changes requested):
Well done @asafgabai!

src/main/java/com/jfrog/ide/idea/ui/LocalComponentsTree.java
asafgabai: It doesn't work because root.children() is Enumeration, so I found another solution:
```java
//noinspection unchecked
cache.cacheNodes((List<FileTreeNode>)(List<?>) Collections.list(root.children()));
```

PR #297, opened 2023-02-22
Welcome and empty state displays

asafgabai:
-----
image
image
image
image


src/main/java/com/jfrog/ide/idea/configuration/GlobalSettings.java
```diff
@@ -167,19 +186,18 @@ public boolean areArtifactoryCredentialsSet() {
...
-     * @return true if connection details from CLI were loaded.
      */
-    public boolean loadConnectionDetailsFromJfrogCli() {
+    public void loadConnectionDetailsFromJfrogCli() {
         try {
             if (serverConfig.readConnectionDetailsFromJfrogCli()) {
                 Logger.getInstance().info("Successfully loaded config connection details from JFrog CLI");
-                return true;
+                MessageBus messageBus = ApplicationManager.getApplication().getMessageBus();
+                messageBus.syncPublisher(ApplicationEvents.ON_CONFIGURATION_DETAILS_CHANGE).update();
```
yahavi, line 195: Shouldn't we remove these lines?
Because reloadXrayCredentials sets the credentials from JFrog CLI and returns true

yahavi (changes requested):

PR #298, opened 2023-02-22
Fix analyzer manager download

talarian1:
-----
This PR fixes the checksum calculation of the remote AnalyzerManager.zip.
As well as handles the analyzerManger logs.

### DIFF
--- a/src/main/java/com/jfrog/ide/idea/scan/ScanBinaryExecutor.java
+++ b/src/main/java/com/jfrog/ide/idea/scan/ScanBinaryExecutor.java
@@ -21,6 +21,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.SystemUtils;
 import org.apache.http.Header;
 import org.jfrog.build.api.util.Log;
+import org.jfrog.build.api.util.NullLog;
 import org.jfrog.build.client.ProxyConfiguration;
 import org.jfrog.build.extractor.clientConfiguration.ArtifactoryManagerBuilder;
 import org.jfrog.build.extractor.clientConfiguration.client.artifactory.ArtifactoryManager;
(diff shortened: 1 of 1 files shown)


src/main/java/com/jfrog/ide/idea/scan/ScanBinaryExecutor.java
```diff
@@ -155,9 +159,9 @@ private void updateBinaryIfNeeded() throws IOException {
 
     public String getFileChecksumFromServer() throws IOException {
         try (ArtifactoryManager artifactoryManager = artifactoryManagerBuilder.build()) {
-            Header[] headers = artifactoryManager.downloadHeaders(JFROG_RELEASES + getBinaryDownloadURL());
+            Header[] headers = artifactoryManager.downloadHeaders(getBinaryDownloadURL());
             for (Header header : headers) {
-                if (header.getName().equals("x-checksum-sha256")) {
+                if (header.getName().toLowerCase().equals("x-checksum-sha256")) {
```
yahavi, line 164:
```suggestion
                if (StringUtils.equalsIgnoreCase(header.getName(), "x-checksum-sha256")) {
```

src/main/java/com/jfrog/ide/idea/scan/ScanBinaryExecutor.java
```diff
@@ -115,7 +118,8 @@ protected List<JFrogSecurityWarning> execute(ScanConfig.Builder inputFileBuilder
 
             Logger log = Logger.getInstance();
             // Execute the external process
-            CommandResults commandResults = commandExecutor.exeCommand(binaryTargetPath.toFile().getParentFile(), args, null, log);
+            log.debug(String.format("Executing command: %s %s", binaryTargetPath.toString(), join(" ", args)));
+            CommandResults commandResults = commandExecutor.exeCommand(binaryTargetPath.toFile().getParentFile(), args, null, new NullLog());
```
yahavi, line 122: Let's add a comment explaining why we send the NullLog and not the logger

yahavi (changes requested):

PR #306, opened 2023-02-24
Fix applicability scan exclude directories 

talarian1:
-----
Add the missing "*/**" to the Applicability skip folders pattern.

### DIFF
--- a/src/main/java/com/jfrog/ide/idea/ui/configuration/JFrogGlobalConfiguration.form
+++ b/src/main/java/com/jfrog/ide/idea/ui/configuration/JFrogGlobalConfiguration.form
@@ -652,7 +652,7 @@
             </constraints>
             <properties>
               <enabled value="false"/>
-              <text value="    Which is used to exclude specific local paths from being scanned by JFrog Xray."/>
+              <text value="    Which is used to exclude specific local directories paths from being scanned by the plugin."/>
             </properties>
           </component>
         </children>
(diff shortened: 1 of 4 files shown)


src/main/java/com/jfrog/ide/idea/ui/configuration/JFrogGlobalConfiguration.form
```diff
@@ -506,7 +506,7 @@
               </grid>
             </constraints>
             <properties>
-              <text value="    Excluded paths"/>
+              <text value="    Excluded directories"/>
               <toolTipText value="Pattern of project paths to exclude from Xray scanning for npm and Go projects."/>
```
asafgabai, line 510: It's a good opportunity to update this tooltip (or even remove it).

src/test/java/com/jfrog/ide/idea/scan/SourceCodeManagerTest.java
```diff
@@ -0,0 +1,39 @@
...
+@RunWith(Parameterized.class)
+public class SourceCodeManagerTest {
+
+    @Parameterized.Parameters
+    public static Collection<Object[]> data() {
+        return Arrays.asList(new Object[][]{
+                {"", new String[]{}},
+                {"**/*{.idea}*", new String[]{"**/*.idea*/**"}},
+                {"**/*.idea*", new String[]{"**/*.idea*/**"}},
+                {"**/*{.idea,test,node_modules}*", new String[]{"**/*.idea*/**", "**/*test*/**", "**/*node_modules*/**"}},
```
asafgabai, line 22: It might be a good idea (if possible) to also test invalid patterns.

asafgabai (changes requested):

src/test/java/com/jfrog/ide/idea/scan/SourceCodeManagerTest.java
talarian1: We already test invalid patterns as part of the ConfigVerificationUtilsNegativeTest :)

PR #309, opened 2023-03-02
Handle JCEF is not supported error

asafgabai:
-----
In Android Studio and older versions of IntelliJ-based IDEs, the plugin was not loaded and the following error was received:
```
JCEF is not supported in this env or failed to initialize
```
The error occurred because in newer versions (2020.2 and later) of most JetBrains IDEs, the boot runtime contains JCEF by default, but it's not the case in Android Studio.
If JCEF is not supported, a message with a suggestion to choose another boot runtime will be shown.

image

In addition, this PR fixes another bug regarding opening links inside the webview, instead of in a new browser window.


src/main/java/com/jfrog/ide/idea/ui/JFrogLocalToolWindow.java
```diff
@@ -159,34 +151,14 @@ private void setLeftPanelContent(JComponent component) {
...
-        }, browser.getCefBrowser());
+    private CefBrowser initVulnerabilityInfoBrowser() throws Exception {
+        tempDirPath = Files.createTempDirectory("jfrog-idea-plugin");
+        Utils.extractFromResources("/jfrog-ide-webview", tempDirPath);
 
         String pageUri = tempDirPath.resolve("index.html").toFile().toURI().toString();
-        browser.loadURL(pageUri);
+        WebviewService webviewService = WebviewService.getInstance();
+        webviewService.setOnLoadEnd(() -> updateIssueOrLicenseInWebview(selectedIssue));
+        return webviewService.initBrowser(pageUri);
```
yahavi, line 161: Can we register a disposer here? I couldn't find any place you did it:
```java
Disposer.register(this, webviewService);
```

src/main/java/com/jfrog/ide/idea/ui/webview/WebviewService.java
```diff
@@ -0,0 +1,87 @@
...
+            public void onLoadError(CefBrowser browser, CefFrame frame, ErrorCode errorCode, String errorText, String failedUrl) {
+                super.onLoadError(browser, frame, errorCode, errorText, failedUrl);
+                Logger.getInstance().error("An error occurred while opening the issue details view: " + errorText);
+            }
+        });
+    }
+
+    public void setOnLoadEnd(Runnable onLoadEnd) {
+        this.onLoadEnd = onLoadEnd;
+    }
```
yahavi, line 81: Looks not in use

yahavi (changes requested):
Important - Collapsed in my IDEA sandbox (not in the latest version)

src/main/java/com/jfrog/ide/idea/ui/webview/WebviewService.java
asafgabai: Used in JFrogLocalToolWindow.initVulnerabilityInfoBrowser().

src/main/java/com/jfrog/ide/idea/ui/JFrogLocalToolWindow.java
```diff
@@ -138,56 +135,67 @@ private void refreshView() {
...
+        // "The plugin uses a component named JCEF that seem to be missing in your IDE."
+        HyperlinkLabel pluginNeedsJcefLabel = new HyperlinkLabel();
+        pluginNeedsJcefLabel.setText("The plugin uses a component named JCEF that seem to be missing in your IDE.");
+        ComponentUtils.addCenteredHyperlinkLabel(jcefNotSupportedPanel, pluginNeedsJcefLabel);
 
-        return ComponentUtils.createUnsupportedPanel(noCredentialsPanel);
+        // "To make JCEF available in your IDE, you’ll need to have the IDE use a different boot runtime."
+        HyperlinkLabel replaceBootRuntimeLabel = new HyperlinkLabel();
+        replaceBootRuntimeLabel.setText("To make JCEF available in your IDE, you’ll need to have the IDE use a different boot runtime.");
+        ComponentUtils.addCenteredHyperlinkLabel(jcefNotSupportedPanel, replaceBootRuntimeLabel);
```
yahavi, line 173: What's the reason for using HyperlinkLabel when the component is no hyperlink?

src/main/java/com/jfrog/ide/idea/ui/JFrogToolWindow.java
```diff
@@ -19,15 +20,22 @@ public class JFrogToolWindow {
 
     void initToolWindow(@NotNull ToolWindow toolWindow, @NotNull Project project, boolean buildsConfigured) {
         ContentManager contentManager = toolWindow.getContentManager();
-        JFrogLocalToolWindow jfrogLocalContent = new JFrogLocalToolWindow(project);
+        JFrogLocalToolWindow jfrogLocalContent = null;
+        try {
+            jfrogLocalContent = new JFrogLocalToolWindow(project);
+        } catch (Exception e) {
+            Logger.getInstance().error("Local view could not be initialized.", e);
```
yahavi, line 27: This is troublesome. In that case, we may like to get an exception. This log message is nearly invisible.

src/main/java/com/jfrog/ide/idea/ui/JFrogToolWindow.java
```diff
@@ -19,15 +20,22 @@ public class JFrogToolWindow {
...
+        }
         JFrogCiToolWindow jFrogCiContent = new JFrogCiToolWindow(project, buildsConfigured);
         addContent(contentManager, jfrogLocalContent, jFrogCiContent);
     }
 
     private void addContent(ContentManager contentManager, JFrogLocalToolWindow jfrogLocalContent, JFrogCiToolWindow jfrogBuildsContent) {
         ContentFactory contentFactory = ContentFactory.SERVICE.getInstance();
-        Content localContent = contentFactory.createContent(jfrogLocalContent, "Local", false);
-        contentManager.addContent(localContent);
+        if (jfrogLocalContent != null) {
```
yahavi, line 35: The local view must always be not null

yahavi (changes requested):
Amazing as usual, @asafgabai.
Could you please add a section in the readme about Android Studio, explaining the issue and the solution?

src/main/java/com/jfrog/ide/idea/ui/JFrogToolWindow.java
asafgabai: I added an empty state for this case.

PR #310, opened 2023-03-05
Add expiry of 7 days for scan cache files

talarian1:
-----

### DIFF
--- a/src/main/java/com/jfrog/ide/idea/ui/LocalComponentsTree.java
+++ b/src/main/java/com/jfrog/ide/idea/ui/LocalComponentsTree.java
@@ -39,6 +39,7 @@ public class LocalComponentsTree extends ComponentsTree {
     private static final String SHOW_IN_PROJECT_DESCRIPTOR = "Show direct dependency in project descriptor";
     private static final String NO_ISSUES = "Your project was scanned and we didn't find any security issues.";
     private static final String SCANNING = "Scanning...";
+    private static final long VALID_CACHE_TIME = 7 * 24 * 60 * 60 * 1000; // 7 days
 
     private final ScanCache cache;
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/jfrog/ide/idea/ui/LocalComponentsTree.java
```diff
@@ -39,6 +39,7 @@ public class LocalComponentsTree extends ComponentsTree {
     private static final String SHOW_IN_PROJECT_DESCRIPTOR = "Show direct dependency in project descriptor";
     private static final String NO_ISSUES = "Your project was scanned and we didn't find any security issues.";
     private static final String SCANNING = "Scanning...";
+    private static final long VALID_CACHE_TIME = 7 * 24 * 60 * 60 * 1000; // 7 days
```
yahavi, line 42:
```suggestion
    // Week
    private static final long MAX_SCAN_CACHE_AGE_MILLISECS = 7 * 24 * 60 * 60 * 1000;
```

src/main/java/com/jfrog/ide/idea/ui/LocalComponentsTree.java
```diff
@@ -204,7 +205,11 @@ private void setNodesFromCache() {
     }
 
     public boolean isCacheEmpty() {
-        return cache.getScanCacheObject() == null;
+        return cache.getScanCacheObject() == null || !isCacheValid();
+    }
+
+    private boolean isCacheValid() {
+        return System.currentTimeMillis() - cache.getScanCacheObject().getScanTimestamp() < VALID_CACHE_TIME;
```
yahavi, line 212: Should we also delete the old cache in that scenario?

yahavi (approved):

PR #326, opened 2023-03-28
Plugin's Toolbar Modifications

talarian1:
-----
This PR introduces the following changes to the plugin's toolbar:
- Add the last scan time label.
- Add `go to setting` button.

Screenshot 2023-03-28 at 17 14 56

- Disable the `run scan` button while scanning.

Screenshot 2023-03-28 at 17 25 04


src/main/java/com/jfrog/ide/idea/actions/GoToSettingsAction.java
```diff
@@ -0,0 +1,21 @@
...
+import com.intellij.openapi.actionSystem.AnActionEvent;
+import com.intellij.openapi.options.ShowSettingsUtil;
+import com.jfrog.ide.idea.ui.configuration.JFrogGlobalConfiguration;
+
+/**
+ * Created by tala on 9/3/23.
+ */
+public class GoToSettingsAction extends AnAction {
+    public GoToSettingsAction() {
+        super("JFrog Global Configuration", "Go to JFrog global configuration ", AllIcons.General.Settings);
```
asafgabai, line 14: I guess the space after "configuration " can be removed 🙈

src/main/resources/META-INF/plugin.xml
```diff
@@ -77,10 +77,10 @@
 
     <actions>
         <!--suppress PluginXmlCapitalization -->
-        <action id="JFrog.RefreshLocal"
-                class="com.jfrog.ide.idea.actions.RefreshLocalAction"
+        <action id="JFrog.StartLocalScan"
+                class="com.jfrog.ide.idea.actions.StartLocalScanAction"
                 text="Run Scan"
-                description="Force Xray scan"
+                description="Run JFrog scan"
```
asafgabai, line 83: I think it'd be better to just remove this description. What do you think?

asafgabai (changes requested):

PR #343, opened 2023-05-09
Read webview resources directly from the JAR

asafgabai:
-----
Read web resources directly from the JAR, without copying them to a temporary directory first.
Stream the browser's console messages to the plugin's logger.

### DIFF
--- a/src/main/java/com/jfrog/ide/idea/ui/webview/WebviewManager.java
+++ b/src/main/java/com/jfrog/ide/idea/ui/webview/WebviewManager.java
@@ -53,7 +53,11 @@ public class WebviewManager implements Disposable {
         jbCefBrowser.getJBCefClient().addDisplayHandler(new CefDisplayHandlerAdapter() {
             @Override
             public boolean onConsoleMessage(CefBrowser browser, CefSettings.LogSeverity level, String message, String source, int line) {
-                Logger.getInstance().info(String.format("Webview console message: %s - %s", level, message));
+                if (level == CefSettings.LogSeverity.LOGSEVERITY_VERBOSE) {
+                    Logger.getInstance().debug(String.format("Webview console message: %s - %s", level, message));
+                } else {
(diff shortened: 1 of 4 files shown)


src/main/java/com/jfrog/ide/idea/ui/webview/WebviewManager.java
```diff
@@ -43,6 +49,16 @@ public void onLoadError(CefBrowser browser, CefFrame frame, ErrorCode errorCode,
         }, jbCefBrowser.getCefBrowser());
     }
 
+    private void streamConsoleMessagesToLog() {
+        jbCefBrowser.getJBCefClient().addDisplayHandler(new CefDisplayHandlerAdapter() {
+            @Override
+            public boolean onConsoleMessage(CefBrowser browser, CefSettings.LogSeverity level, String message, String source, int line) {
```
talarian1, line 55: Should we also consider the `CefSettings.LogSeverity` in our logger?
For instance, logging `LOGSEVERITY_VERBOSE` only in debug?

src/main/java/com/jfrog/ide/idea/ui/webview/WebviewResourceHandler.java
```diff
@@ -0,0 +1,109 @@
...
+            connection = currUrl.openConnection();
+            if (inputStream != null) {
+                inputStream.close();
+            }
+            inputStream = connection.getInputStream();
+            error = false;
+            callback.Continue();
+            return true;
+        } catch (IOException | NullPointerException e) {
+            Logger.getInstance().error("An error occurred while reading webview resource.", e);
```
talarian1, line 41: Should we also include the resource path (or name) for debugging purposes?

asafgabai: Yes. Good idea, I'll add it.

PR #351, opened 2023-05-22
Add Frogbot scan and fix workflow

EyalDelarea:
-----


EyalDelarea:
Frogbot scan and fix not supported

PR #360, opened 2023-06-19
Add new `Infrastructure as code` and `secrets` scanners

talarian1:
-----


src/main/resources/META-INF/with-java.xml
```diff
@@ -1,5 +1,4 @@
 <idea-plugin>
     <extensions defaultExtensionNs="com.intellij">
-        <externalAnnotator language="JAVA" implementationClass="com.jfrog.ide.idea.inspections.JFrogSecurityAnnotator"/>
     </extensions>
```
asafgabai, line 3: It looks like this file is not needed anymore. Can it be removed?

src/main/resources/META-INF/with-python.xml
```diff
@@ -2,6 +2,5 @@
     <depends>com.intellij.modules.platform</depends>
     <depends>com.intellij.modules.python-pro-capable</depends>
     <extensions defaultExtensionNs="com.intellij">
```
asafgabai, line 4: I guess the "extensions" tag can be removed. Doesn't it?

asafgabai (changes requested):

src/main/java/com/jfrog/ide/idea/scan/ScanBinaryExecutor.java
```diff
@@ -211,14 +223,24 @@ protected boolean shouldExecute() {
...
         List<JFrogSecurityWarning> warnings = new ArrayList<>();
-        output.getRuns().forEach(run -> run.getResults().forEach(result -> warnings.add(new JFrogSecurityWarning(result))));
+        output.getRuns().forEach(run -> run.getResults().forEach(result -> warnings.add(new JFrogSecurityWarning(result, scanType))));
+
+        Optional<Run> run = output.getRuns().stream().findFirst();
+        if (run.isPresent()) {
+            List<Rule> scanners = run.get().getTool().getDriver().getRules();
+            // Adds the scanner search target data
+            for (JFrogSecurityWarning warning : warnings) {
+                String ScannerSearchTarget = scanners.stream().filter(scanner -> scanner.getId().equals(warning.getName())).findFirst().map(Rule::getFullDescription).map(Message::getText).orElse("");
```
asafgabai, line 240: ScannerSearchTarget
-->
scannerSearchTarget

asafgabai (changes requested):

PR #362, opened 2023-06-21
Flat dependency trees

asafgabai:
-----

### DIFF
--- a/src/main/java/com/jfrog/ide/idea/scan/PypiScanner.java
+++ b/src/main/java/com/jfrog/ide/idea/scan/PypiScanner.java
@@ -1,6 +1,5 @@
 package com.jfrog.ide.idea.scan;
 
-import com.google.common.collect.Sets;
 import com.intellij.execution.ExecutionException;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.projectRoots.Sdk;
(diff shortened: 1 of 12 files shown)


settings.gradle
```diff
@@ -7,4 +7,10 @@ pluginManagement {
     }
 }
 
+sourceControl {
```
asafgabai, line 10: Reminder for myself to remove this when it's released.

src/main/java/com/jfrog/ide/idea/scan/PypiScanner.java
```diff
@@ -52,116 +44,57 @@ static List<Sdk> getAllPythonSdks() {
...
+            compIdByCompName.put(pyPackage.getName().toLowerCase(), compId);
+            directDeps.add(compId);
         }
 
-        for (PyRequirement requirement : childPyPackage.getRequirements()) {
-            PyPackage dependency = dependencyMapping.get(requirement.getName().toLowerCase());
-            if (dependency == null) {
-                getLog().warn("Dependency " + requirement.getName() + " is not installed.");
-                continue;
+        Map<String, DepTreeNode> nodes = new HashMap<>();
```
talarian1, line 76: Let's add some comments here :)

talarian1 (changes requested):

PR #365, opened 2023-07-03
Upgrade JFrog-IDE-Webview to 0.2.1

Or-Geva:
-----

### DIFF
--- a/src/main/java/com/jfrog/ide/idea/ui/jcef/message/MessageType.java
+++ b/src/main/java/com/jfrog/ide/idea/ui/jcef/message/MessageType.java
@@ -1,0 +1,16 @@
+package com.jfrog.ide.idea.ui.jcef.message;
+
+public enum MessageType {
+    SetEmitter("SET_EMITTER"),
+    ShowPage("SHOW_PAGE");
+
+    private final String value;
+
+    MessageType(String value) {
+        this.value = value;
+    }
+
+    public String getValue() {
+        return value;
+    }
+}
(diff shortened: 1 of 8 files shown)


src/main/java/com/jfrog/ide/idea/ui/jcef/message/PackedMessage.java
```diff
@@ -3,25 +3,31 @@
 import java.io.Serializable;
 
 public class PackedMessage implements Serializable {
+    private Object pageData;
+    private String type;
 
-    public static final String BROWSER_SEND_FUNCTION_NAME = "sendMessageToBrowser";
-    public static final String IDE_SEND_FUNCTION_NAME = "sendMessageToIde";
+    public PackedMessage() {
+        this.pageData = null;
```
talarian1, line 10: I'm not sure this line is needed.

src/main/java/com/jfrog/ide/idea/ui/jcef/message/MessageType.java
```diff
@@ -0,0 +1,16 @@
+package com.jfrog.ide.idea.ui.jcef.message;
+
+public enum MessageType {
+    SetEmitter("SET_EMITTER"),
```
talarian1, line 4:
```suggestion
    SET_EMITTER, SHOW_PAGE
```
https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html
I think this will be easier to read and follows Java conventions :)

talarian1 (changes requested):

PR #368, opened 2023-07-12
Quick fix UX improvements

asafgabai:
-----
UX improvements regarding the dependency upgrade process (quick fix):
- show a progress bar (indicator).
- refresh the file when the upgrade is done.

Bug fix - in Go, npm and Yarn projects, upgrading dependencies worked on the project's root directory only.


src/main/java/com/jfrog/ide/idea/inspections/upgradeversion/UpgradeVersion.java
```diff
@@ -50,11 +52,19 @@ public String getFamilyName() {
...
-        }
+        Task.Backgroundable scanAndUpdateTask = new Task.Backgroundable(project, "Upgrading dependency...") {
+            @Override
+            public void run(@NotNull com.intellij.openapi.progress.ProgressIndicator indicator) {
+                try {
+                    upgradeComponentVersion(project, descriptor);
+                    ApplicationManager.getApplication().invokeLater(() -> descriptor.getPsiElement().getContainingFile().getVirtualFile().refresh(false, false));
+                    log.info("Upgraded " + componentName + " to version " + fixVersion + " successfully.");
+                } catch (Exception e) {
+                    log.error("Failed while trying to upgrade component version. Error: " + e);
```
talarian1, line 63: Can we add the component name and version? (If it is not already in the error message itself)

src/main/java/com/jfrog/ide/idea/scan/ScannerBase.java
```diff
@@ -219,7 +219,7 @@ private void visitDepTreeNode(Map<String, DependencyNode> dependencies, DepTree
                 }
                 DependencyNode existingDep = addedDeps.get(descriptorPath).get(compId);
                 if (existingDep != null) {
-                    existingDep.setIndirect(indirect);
+                    existingDep.setIndirect(indirect && existingDep.isIndirect());
```
talarian1, line 222:
```suggestion
                   // Preserve previous "direct" information  
                    if(existingDep.isIndirect()){
                       existingDep.setIndirect(indirect);
                     }
```
I believe it is a bit more clear that way :)

talarian1 (changes requested):

PR #370, opened 2023-07-16
Webview - Add event message passing from the Webview to the IDE

Or-Geva:
-----

This pull request introduces the capability to capture event notifications from the JFrog IDE Webview.
It includes the initial event type called JumpToCode, which will be accessible on the Eos view. Clicking on a specific button on the Eos view will now open the corresponding file and highlight the relevant line based on the Webview.

Additionally, the entire event mechanism has been refactored for improved functionality.

Depend on:
* https://github.com/jfrog/jfrog-ide-webview/pull/26


src/main/java/com/jfrog/ide/idea/ui/webview/event/model/IdeEvent.java
```diff
@@ -0,0 +1,22 @@
...
+
+    public Type getType() {
+        return type;
+    }
+
+    public void setType(Type type) {
+        this.type = type;
+    }
+
+    public enum Type {SHOW_CODE}
```
asafgabai, line 21: If it's not too complicated, I recommend calling this type "JUMP_TO_CODE" to make it consistent.

src/main/java/com/jfrog/ide/idea/ui/webview/event/model/IdeEvent.java
```diff
@@ -0,0 +1,22 @@
+package com.jfrog.ide.idea.ui.webview.event.model;
+
+/**
+ * Represents an IDE-specific event that can be sent from the IDE to the Webview.
```
asafgabai, line 4: Isn't it sent from the webview to the IDE?
It seems to be the other way around also in WebviewEvent.
Or did I just get it wrong?

src/main/java/com/jfrog/ide/idea/ui/webview/event/Sender.java
```diff
@@ -0,0 +1,77 @@
...
+    /**
+     * Sends a webview event with the specified type and data to the webview.
+     *
+     * @param type The type of the webview event.
+     * @param data The data associated with the webview event.
+     */
+    public void sendEvent(WebviewEvent.Type type, Object data) {
+        try {
+            String raw = pack(type, data);
+            Logger.getInstance().debug("Sending data to jfrog  webview: " + raw);
```
asafgabai, line 62: jfrog _ _ webview
-->
jfrog _ webview
🫣

src/main/java/com/jfrog/ide/idea/ui/webview/event/tasks/JumpToCodeTask.java
```diff
@@ -0,0 +1,22 @@
...
+import com.intellij.openapi.project.Project;
+import com.jfrog.ide.idea.inspections.JumpToCode;
+import com.jfrog.ide.idea.ui.webview.model.Location;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Represents a task that performs the "Jump to Code" action in the IDE.
+ * This task is responsible for executing the jump to code operation based on the provided project and location.
+ */
+public class JumpToCodeTask {
```
asafgabai, line 12: What's the purpose of this class? Why don't you just use JumpToCode directly instead?

asafgabai (changes requested):

src/main/java/com/jfrog/ide/idea/inspections/JumpToCode.java
```diff
@@ -0,0 +1,97 @@
...
+     * @param endRow      The ending row of the code range.
+     * @param startColumn The starting column of the code range.
+     * @param endColumn   The ending column of the code range.
+     */
+    public void execute(String filePath, int startRow, int endRow, int startColumn, int endColumn) {
+        if (this.project == null || this.fileEditorManager == null) return;
+        VirtualFile file = getVirtualFile(filePath);
+        if (file == null) return;
+        ApplicationManager.getApplication().invokeLater(() -> {
+            ApplicationManager.getApplication().invokeAndWait(() -> {
```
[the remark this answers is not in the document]
Or-Geva: the `ApplicationManager.invokeLater()` method is used to wrap the code that interacts with the IntelliJ IDEA API. This ensures that the code runs on the dispatch thread.
`invokeAndWait()` method is used within `invokeLater()` to ensure that the code completes synchronously before moving on.

src/main/java/com/jfrog/ide/idea/ui/webview/event/EventManager.java
```diff
@@ -0,0 +1,47 @@
...
+import com.intellij.ui.jcef.JBCefBrowser;
+import com.jfrog.ide.idea.ui.webview.event.model.WebviewEvent;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * The EventManager responsible for managing events between the IDE and the Webview.
+ * It handles the creation of a receiver and sender, allowing communication between the components.
+ */
+public class EventManager {
+    public static String ideSendFuncName = "sendMessageToIdeFunc";
```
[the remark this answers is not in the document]
Or-Geva: Yes

src/main/java/com/jfrog/ide/idea/ui/webview/event/Receiver.java
```diff
@@ -0,0 +1,84 @@
...
+
+    /**
+     * Handles the received IdeEvent.
+     *
+     * @param event The received IdeEvent to handle.
+     */
+    private void handler(IdeEvent event) {
+        if (Objects.requireNonNull(event.getType()) == IdeEvent.Type.SHOW_CODE) {
+            new JumpToCodeTask(this.project).execute(createMapper().convertValue(event.getData(), Location.class));
+            Logger.getInstance().error("Jump to " + event.getType());
```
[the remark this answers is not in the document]
Or-Geva: Yea, thanks!

src/main/java/com/jfrog/ide/idea/ui/webview/event/model/IdeEvent.java
Or-Geva: You are correct, i will switch between both ->
```
 Represents a Webview-specific event that can be sent from the IDE to the Webview.
```

src/main/java/com/jfrog/ide/idea/ui/webview/event/tasks/JumpToCodeTask.java
Or-Geva: The purpose behind this is to serve as a _guideline_ for upcoming tasks. In my perspective, having a `handler(IdeEvent event)` would function as the central starting point for various tasks. However, the intention is not for every task to embed its logic within the handler function. Instead, the idea is to have each task encapsulated within its own class, containing the appropriate logic.

src/main/java/com/jfrog/ide/idea/inspections/JumpToCode.java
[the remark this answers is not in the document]
asafgabai: According to these methods' docs, it seems that both are almost the same.
The difference between them is that invokeLater runs the Runnable asynchronously, and invokeAndWait blocks until the Runnable returns.
In your code, it looks like you don't wait for anything, as there are no more lines in the invokeLater block.
So I guess just using invokeLater (without invokeAndWait inside) would bring us the same result.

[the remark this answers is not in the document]
Or-Geva: Ok, I've removed `invokeAndWait()`

PR #374, opened 2023-07-18
Update Analyzer Manager version to 1.2.2.1846078

talarian1:
-----
This PR upgrade Analyzer Manager version that supports Secrats and Iac scan as well as contextual analysis for Java projects. 
Update integration test accordingly.


src/test/java/com/jfrog/ide/idea/integration/ApplicabilityScannerIntegrationTests.java
```diff
@@ -81,6 +81,23 @@ public void testApplicabilityScannerPythonProject() throws IOException, Interrup
         assertTrue(results.get(0).getFilePath().endsWith("main.py"));
     }
 
+    public void testApplicabilityScannerMavenProject() throws IOException, InterruptedException {
```
asafgabai, line 84: I suggest changing the test name to testApplicabilityScannerJavaProject to make it clear that it's not specific to Maven, but it covers Java projects in general.

asafgabai (changes requested):

PR #383, opened 2023-08-01
Limit Impact Graph size

asafgabai:
-----
Limit the Impact Graph size to up to 10 paths, to avoid high memory consumption and improve readability.


talarian1 (approved):

src/main/java/com/jfrog/ide/idea/scan/ScannerBase.java
```diff
@@ -58,6 +59,8 @@
  * Created by romang on 4/26/17.
  */
 public abstract class ScannerBase {
+    public static final int IMPACT_PATHS_LIMIT = 10;
```
Or-Geva, line 62:
```suggestion
    public static final int IMPACT_PATHS_LIMIT = 50;
```

PR #391, opened 2023-08-15
Show floating toolbar when results don't match source code

talarian1:
-----
When the source code no longer matches the scan results, suggest re-scan by showing our `JFrog floating toolbar`.

Screenshot 2023-08-15 at 12 59 31

### DIFF
--- a/src/main/java/com/jfrog/ide/idea/ui/JFrogFloatingToolbar.java
+++ b/src/main/java/com/jfrog/ide/idea/ui/JFrogFloatingToolbar.java
@@ -13,7 +13,9 @@ import com.jfrog.ide.idea.events.ApplicationEvents;
 import com.jfrog.ide.idea.utils.Descriptor;
 import org.jetbrains.annotations.NotNull;
 
+import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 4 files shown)


src/main/java/com/jfrog/ide/idea/ui/JFrogFloatingToolbar.java
```diff
@@ -31,18 +45,51 @@ public void register(@NotNull DataContext dataContext, @NotNull FloatingToolbarC
...
+        projectBusConnection.subscribe(ApplicationEvents.ON_SCAN_LOCAL_STARTED, (ApplicationEvents) this::clear);
+    }
+
+    private void clear() {
+        this.changedFiles = new HashSet<>();
+        filesToolbarComponents.values().forEach(FloatingToolbarComponent::scheduleHide);
+        filesToolbarComponents = new HashMap<>();
+    }
+
+    private void updateFileChanged(String s) {
```
asafgabai, line 82: I would suggest renaming `s` to `filePath` or something similar.

asafgabai (changes requested):

PR #408, opened 2023-08-24
Consider all Python packages as direct dependencies 

talarian1:
-----
As a result of the way Python packages are consumed by Python projects, we can't confidently say if a package is a direct dependency of the current project. It was decided that is more accurate to treat all Python dependencies as direct dependencies and include them all in the Contextual Analysis scan.
Please notice that the `Populate each node's children ` section is still needed for the impact graph display:

Screenshot 2023-08-24 at 18 40 15

### DIFF
--- a/src/main/java/com/jfrog/ide/idea/scan/PypiScanner.java
+++ b/src/main/java/com/jfrog/ide/idea/scan/PypiScanner.java
@@ -86,7 +86,6 @@ public class PypiScanner extends SingleDescriptorScanner {
                     continue;
                 }
                 node.getChildren().add(depId);
-                directDeps.remove(depId);
             }
             nodes.put(compId, node);
         }


src/main/java/com/jfrog/ide/idea/scan/PypiScanner.java
```diff
@@ -86,7 +86,6 @@ private DepTree createSdkDependencyTree(Sdk pythonSdk) throws ExecutionException
                     continue;
                 }
                 node.getChildren().add(depId);
-                directDeps.remove(depId);
```
asafgabai, line 89: I'd suggest changing the name of `directDeps` to something else. Maybe `dependencies`?
Also, the comment in line 69 needs to be updated with the new logic. It'd be best to explain why we treat all of the dependencies as direct.

asafgabai (changes requested):

PR #415, opened 2023-09-11
Improve source code scanners logging

talarian1:
-----
The scanner's run logs will now be logged in the IDE log instead of in a dedicated log file.
If an error occurs the log level is `INFO`, otherwise `DEBUG`.


src/main/java/com/jfrog/ide/idea/scan/ScanBinaryExecutor.java
```diff
@@ -158,18 +157,26 @@ protected List<JFrogSecurityWarning> execute(ScanConfig.Builder inputFileBuilder
...
-                log.debug(String.format("Scanner %s is not supported in the current Analyzer Manager version.", scanType));
-                return List.of();
-            }
-            if (!commandResults.isOk()) {
-                throw new IOException(commandResults.getErr());
+
+            if (commandResults.isOk()) {
+                log.debug(commandResults.getRes());
+                return parseOutputSarif(outputFilePath);
+            } else {
```
yahavi, line 164: Let's remove the "else" after return

yahavi (approved):

yahavi:
I think this change should be considered an "improvement" and not "ignore for release".

PR #426, opened 2023-10-04
Fix creating multiple nodes of the same file

asafgabai:
-----
Fix creating multiple nodes of the same file.


src/main/java/com/jfrog/ide/idea/ui/LocalComponentsTree.java
```diff
@@ -63,22 +63,27 @@ public static LocalComponentsTree getInstance(@NotNull Project project) {
...
-
-    private void appendProject(FileTreeNode filteredRoot) {
-        SortableChildrenTreeNode root;
-        if (getModel() == null) {
-            root = new SortableChildrenTreeNode();
-        } else {
-            root = (SortableChildrenTreeNode) getModel().getRoot();
-        }
+        ApplicationManager.getApplication().invokeLater(() -> {
+            SortableChildrenTreeNode root;
```
talarian1, line 67:
```suggestion
        SortableChildrenTreeNode root = getModel() != null ?(SortableChildrenTreeNode) getModel().getRoot(): new SortableChildrenTreeNode();
```

src/main/java/com/jfrog/ide/idea/ui/LocalComponentsTree.java
```diff
@@ -63,22 +63,27 @@ public static LocalComponentsTree getInstance(@NotNull Project project) {
...
+                root = new SortableChildrenTreeNode();
+            } else {
+                root = (SortableChildrenTreeNode) getModel().getRoot();
+            }
 
-        root.add(filteredRoot);
-        root.sortChildren();
-        populateTree(root);
+            for (FileTreeNode node : fileTreeNodes) {
+                if (root.getChildren() != null) {
```
talarian1, line 75: What do you think about using:
 ```java
Optional.ofNullable(root.getChildren())
                .orElseGet(Collections::emptyList)
```
 instead of the:
```java
if (root.getChildren() != null) {...}
```
?

src/test/java/com/jfrog/ide/idea/ui/LocalComponentsTreeTest.java
```diff
@@ -0,0 +1,44 @@
...
+        );
+
+        ApplicationManager.getApplication().invokeAndWait(() -> tree.doAddScanResults(nodes1));
+        assertEquals(2, getTreeFileNodes(tree).size());
+        List<FileTreeNode> nodes2 = List.of(
+                createFileTreeNodeWithIssue("path/to/file1.txt", "issue3"),
+                createFileTreeNodeWithIssue("path/to/file3.txt", "issue4")
+        );
+        ApplicationManager.getApplication().invokeAndWait(() -> tree.doAddScanResults(nodes2));
+        assertEquals(3, getTreeFileNodes(tree).size());
```
talarian1, line 28: Can we also assert here that the node `path/to/file1.txt` has 2 issues ?

PR #429, opened 2023-10-11
Add Abort Scan button 

talarian1:
-----
This PR adds a new abort all scan's task option.
stop


src/main/java/com/jfrog/ide/idea/scan/ScannerBase.java
```diff
@@ -124,7 +125,8 @@ void setScanLogic(ScanLogic logic) {
      *
      * @return the Inspection tool corresponding to the scan-manager type.
      */
-    protected @Nullable abstract AbstractInspection getInspectionTool();
+    protected @Nullable
+    abstract AbstractInspection getInspectionTool();
```
asafgabai, line 129: Let's put it back in the same line 🙂

asafgabai (changes requested):

PR #431, opened 2023-10-17
Fix compatibility problems in future IntelliJ releases

asafgabai:
-----
Use the Apache Commons Lang dependency directly, instead of calling the one used by IntelliJ (it's already directly imported, but was used only in part of its usages).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,7 +9,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu-latest, windows-latest, macOS-latest ]
-        testCommand: [ "verifyPlugin runPluginVerifier test", "pythonTests" ]
+        testCommand: [ "test", "pythonTests" ]
         include:
           - os: windows-latest
             gradlewSuffix: .bat
--- a/src/main/resources/META-INF/plugin.xml
+++ b/src/main/resources/META-INF/plugin.xml
@@ -36,7 +36,6 @@
         ]]>
     </change-notes>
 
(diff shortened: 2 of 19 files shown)


Or-Geva (approved):

.github/workflows/test.yml
```diff
@@ -9,7 +9,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu-latest, windows-latest, macOS-latest ]
-        testCommand: [ "verifyPlugin runPluginVerifier test", "pythonTests" ]
+        testCommand: [ "test", "pythonTests" ]
```
Or-Geva, line 12: Let's open a new PR to fix python tests

src/main/resources/META-INF/plugin.xml
```diff
@@ -36,7 +36,6 @@
         ]]>
     </change-notes>
 
-    <idea-version since-build="223.4884.69"/>
```
Or-Geva, line 39: is it on purpose?

Or-Geva (changes requested):

src/main/resources/META-INF/plugin.xml
asafgabai: Yes. I moved the sinceBuild property to the build.gradle file instead. It's needed for the verification task, as a minimum version to verify against. When the plugin is being built, this property is injected here automatically (for that to happen, I removed the `updateSinceUntilBuild = false` line from the build.gradle file).

PR #441, opened 2023-11-23
Optimized impact paths building

asafgabai:
-----
Depends on https://github.com/jfrog/jfrog-ide-webview/pull/49

Separated the process of grouping dependencies in file nodes from the process of building impact paths.
Improved both processes for better performance.
Impact paths are built bottom-up (from the vulnerable dependency to the root of the project) to avoid scanning the whole dependency tree.
Dependencies of Maven projects are grouped in file nodes in a similar, more efficient way, to find out which modules depend on vulnerable dependencies, without the need to find all possible paths between these dependencies and the roots of all modules.


src/main/java/com/jfrog/ide/idea/ui/webview/WebviewObjectConverter.java
```diff
@@ -143,7 +143,7 @@ public static DependencyPage convertLicenseToDepPage(LicenseViolationNode licens
     }
 
     private static ImpactGraph convertImpactGraph(ImpactTree impactTree) {
-        return new ImpactGraph(convertImpactGraphNode(impactTree.getRoot()), impactTree.getImpactPathsCount(), ScannerBase.IMPACT_PATHS_LIMIT);
+        return new ImpactGraph(convertImpactGraphNode(impactTree.getRoot()), impactTree.getImpactPathsCount() >= ImpactTreeBuilder.IMPACT_PATHS_LIMIT ? impactTree.getImpactPathsCount() : -1);
```
asafgabai, line 146: Reminder for myself to upgrade the Webview version before merging.

src/main/java/com/jfrog/ide/idea/scan/utils/ImpactTreeBuilder.java
```diff
@@ -0,0 +1,78 @@
...
+    /**
+     * Builds impact paths for {@link DependencyNode} objects.
+     *
+     * @param vulnerableDependencies a map of component IDs and the DependencyNode object matching each of them
+     * @param parents                a map of all dependencies and their parents
+     * @param rootId                 the project's root component ID
+     * @return list of {@link FileTreeNode}s which are all {@link DescriptorFileTreeNode}s with the given
+     * {@link DependencyNode}s inside them
+     */
+    public static void populateImpactTrees(Map<String, DependencyNode> vulnerableDependencies, Map<String, Set<String>> parents, String rootId) {
```
eyalbe4, line 23: I think it's a bit unclear which of this method's argument is actually an output of the method. Maybe we should explain this better as part of the method description, or better, return the output as a return value.

eyalbe4 (changes requested):

src/main/java/com/jfrog/ide/idea/scan/utils/ImpactTreeBuilder.java
asafgabai: I prefer not to return the DependencyNodes to make it clear that the method doesn't create new DependencyNodes but populates the existing ones. I updated the method description to make it clearer.

PR #450, opened 2023-12-11
Introducing Verification Code for SSO Login

noyshabtay:
-----
* Verification code added to SSO login process.

### DIFF
--- a/src/main/java/com/jfrog/ide/idea/ui/configuration/JFrogGlobalConfiguration.form
+++ b/src/main/java/com/jfrog/ide/idea/ui/configuration/JFrogGlobalConfiguration.form
@@ -17,7 +17,7 @@
     </properties>
     <border type="none"/>
     <children>
-      <grid id="b0c38" binding="connectionDetails" layout-manager="GridLayoutManager" row-count="16" column-count="4" same-size-horizontally="false" same-size-vertically="false" hgap="-1" vgap="-1">
+      <grid id="b0c38" binding="connectionDetails" layout-manager="GridLayoutManager" row-count="17" column-count="4" same-size-horizontally="false" same-size-vertically="false" hgap="-1" vgap="-1">
         <margin top="0" left="0" bottom="0" right="0"/>
         <constraints>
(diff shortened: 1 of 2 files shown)


asafgabai (approved):

src/main/java/com/jfrog/ide/idea/ui/configuration/JFrogGlobalConfiguration.form
```diff
@@ -246,10 +246,27 @@
...
           </component>
+          <component id="fd2cc" class="javax.swing.JLabel" binding="ssoLoginInstructionsLabel">
+            <constraints>
+              <grid row="8" column="0" row-span="1" col-span="4" vsize-policy="0" hsize-policy="0" anchor="8" fill="0" indent="0" use-parent-layout="false"/>
+            </constraints>
+            <properties>
+              <text value="SSO login instructions"/>
+            </properties>
+          </component>
+          <component id="a0ab" class="javax.swing.JLabel" binding="ssoCode">
```
yahavi, line 261: Let's make it a non-editable JBTextField to make it copyable.

Also, please consider rearranging it to make it as follows:
image

yahavi (changes requested):
Please make sure that:
1. Nothing moves after clicking on a button
2. The code is copyable

src/main/java/com/jfrog/ide/idea/ui/configuration/JFrogGlobalConfiguration.java
```diff
@@ -486,6 +491,9 @@ private void initLoginViaBrowserButton() {
      */
     private void doSsoLogin() {
         String uuid = UUID.randomUUID().toString();
+        String code = uuid.substring(uuid.length() - 4);
+        ssoCode.setText("Verification code: " + code);
+        ssoLoginInstructionsLabel.setText("After you login using the browser, you’ll be asked to enter the verification code above.");
```
Or-Geva, line 496: Let's have the same text JFrog has:

```
After logging in via your web browser, please enter the code if prompted: " + code]
```

PR #466, opened 2024-01-24
Fix running executables with spaces in their paths in Windows

attiasas:
-----

Update build-info version to 2.41.13

JAS scanners are not running on Windows machines where the user name has space in there name.
Updating build-info version to apply the fix for this issue: https://github.com/jfrog/build-info/pull/777

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -48,7 +48,7 @@ repositories {
     }
 }
 
-def buildInfoVersion = '2.41.9'
+def buildInfoVersion = '2.41.13'
 def idePluginsCommonVersion = '2.3.5'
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -48,7 +48,7 @@ repositories {
     }
 }
 
-def buildInfoVersion = '2.41.9'
+def buildInfoVersion = '2.41.13'
```
asafgabai, line 51: Let's update the build-info version in ide-plugins-common too.

asafgabai (changes requested):

PR #475, opened 2024-08-27
Feature/add support for jetbrains ides

eyalk007:
- [ X] All [tests](https://github.com/jfrog/jfrog-idea-plugin/actions/workflows/test.yml) passed. If this feature is not already covered by the tests, I added new tests.
-----


eranturgeman:
Change the label to 'new feature' as it is a new ability

hadarshjfrog (approved):

eranturgeman:
Please do not combine several tasks in one PR, even if it will make you open 3 different PRs with a single line each.
You fixed a bug, added new feature and updated version in a yml file. these are 3 different and unrelated tasks.
Dont forget this is an OPEN SOURCE project. Every Pr must be clear what it contains and what it changes, and contain only changes related to a single task (think what will happen if we need to revert this commit - all of your work will be lost even if only 1 thing from this PR caused an issue).
This is also important to reflect our work correctly in the release notes. This PR will go under "bug fixes" even when it contains a new ability, of will go under "new features" when it suggest a fix for an issue, that customers might be miss this fix

eranturgeman (changes requested):
refer to my comments

eranturgeman:
Please write in detail exactly what ides you added support to. If you didn't check all ides (what does it even mean? there are plenty of IDES we doesn't support at all) so we cannot make this commitment. If you did check them one by one please list their names

eranturgeman (approved):
Great work!

PR #477, opened 2024-09-01
Upgraded the AM version

eyalk007:
-----

Upgraded the Analyzer Manager version.
The AM executes all of our security scans, and returns us the vulnerabilities that were found by the scanners.

The upgrade makes it so the used version of AM uses the new JAS engines, CA scanners-as-data, and new SAST scanners.


hadarshjfrog (approved):

build.gradle
```diff
@@ -4,7 +4,7 @@ import java.net.http.HttpResponse
 import java.nio.file.Paths
 
 plugins {
-    id "org.jetbrains.intellij" version "1.16.0"
+    id "org.jetbrains.intellij" version "1.17.0"
```
eyalk007, line 7: this is an experiment for something, will remove before merge

src/main/java/com/jfrog/ide/idea/inspections/JFrogSecurityWarning.java
```diff
@@ -62,7 +62,7 @@ public JFrogSecurityWarning(SarifResult result, SourceCodeScanType reporter) {
                 result.getRuleId(),
                 getFirstRegion(result).getSnippet().getText(),
                 reporter,
-                !result.getKind().equals("pass"),
+                (!result.getKind().equals("pass") && (rule.getRuleProperties().map(properties -> properties.getApplicability().equals("applicable")).orElse(true))),
```
eyalk007, line 65: On one side, this is very long
problem is this is the super, meaning I cant use a private func and call it. 
I can put a defualt value and edit, but that is gross

src/main/java/com/jfrog/ide/idea/scan/data/Rule.java
```diff
@@ -15,6 +16,9 @@ public class Rule {
     @JsonProperty("fullDescription")
     private Message fullDescription;
 
+    @JsonProperty("properties")
+    private RuleProperties properties;
```
eyalk007, line 20: not all rules have properties. should the field itself be optional

src/test/resources/sourceCode/applicable_kind_pass_output.sarif
```diff
@@ -0,0 +1,140 @@
+{
```
eyalk007, line 1: just please check i didnt mess things up. the test passes

build.gradle
hadarshjfrog: reminder to remove

src/main/java/com/jfrog/ide/idea/inspections/JFrogSecurityWarning.java
hadarshjfrog: why not creating a value before it that'll calculate it?
having a map, and getting into and a property of a property of a.... 
isn't a best-practice..

And while you're at it - do the same for line 61

src/main/java/com/jfrog/ide/idea/scan/data/Driver.java
```diff
@@ -48,4 +48,12 @@ public boolean equals(Object other) {
         return (Objects.equals(this.name, rhs.name) && (CollectionUtils.isEqualCollection(this.rules, rhs.rules)));
     }
 
+    public Rule getRuleById(String ruleId) {
```
hadarshjfrog, line 51: please add a test to it, and embed a test for it in: `parseOutputSarif` - where this error is the expected result

src/main/java/com/jfrog/ide/idea/scan/data/Rule.java
hadarshjfrog: is that a question? yes - it should be optional...

src/main/java/com/jfrog/ide/idea/inspections/JFrogSecurityWarning.java
eyalk007: um okay.
I still dont love it . but its an okay alternative

src/test/resources/sourceCode/applicable_kind_pass_output.sarif
```diff
@@ -0,0 +1,140 @@
...
+              "fullDescription": {
+                "text": "The scanner checks whether the vulnerable function `pem.Decode` is called.",
+                "markdown": "The scanner checks whether the vulnerable function `pem.Decode` is called."
+              },
+              "shortDescription": {
+                "text": "Scanner for CVE-2020-28502"
+              }
+            },
+            {
+              "id": "CVE-2022-25978",
```
hadarshjfrog, line 23:
```suggestion
              "id": "applic_CVE-2022-25978",
```

src/test/resources/sourceCode/applicable_kind_pass_output.sarif
```diff
@@ -0,0 +1,140 @@
...
+              }
+            },
+            {
+              "id": "applic_CVE-2022-29019",
+              "fullDescription": {
+                "text": "The scanner checks whether the vulnerable function `org.apache.xmlbeans.XmlObject.Factory.parse` is called or an interface that extends `org.apache.xmlbeans.XmlObject` is used.",
+                "markdown": "The scanner checks whether the vulnerable function `org.apache.xmlbeans.XmlObject.Factory.parse` is called or an interface that extends `org.apache.xmlbeans.XmlObject` is used."
+              },
+              "shortDescription": {
+                "text": "Scanner for CVE-2020-5310"
```
hadarshjfrog, line 53: Please align the CVE numbers with the ID

src/test/resources/sourceCode/applicable_kind_pass_output.sarif
```diff
@@ -0,0 +1,140 @@
...
+            "text": "The scanner checks whether the vulnerable function `ansi-regex` is called."
+          },
+          "kind": "pass",
+          "ruleId": "applic_CVE-2021-25878"
+        },
+        {
+          "message": {
+            "text": "The scanner checks whether the vulnerable function `ansi-regex` is called."
+          },
+          "kind": "pass",
```
hadarshjfrog, line 132: add one "kind": "fail".
Also - support undetermined and not covered

hadarshjfrog (changes requested):
Please add more information on the PR description - use the full AM name. also explain what is does (JAS Scanners enablement etc...)
Explain what capabilities it brings (new JAS engines, CA scanners-as-data, new SAST scanners)

src/main/java/com/jfrog/ide/idea/inspections/JFrogSecurityWarning.java
eyalk007: i just made them static instead

src/main/java/com/jfrog/ide/idea/scan/data/Driver.java
eyalk007: i am unsure where to put the test honestly

hadarshjfrog (approved):

src/main/java/com/jfrog/ide/idea/scan/data/Driver.java
eyalk007: check again. was added to sarif.

hadarshjfrog (approved):

PR #479, opened 2024-09-08
Configured the verify plugin command 

eyalk007:
-----

This PR updates the config in the build gradle for verify plugin command.
Instead of checking the plugin on every version of IDEA, and running up the runner space, it now checks only on the oldest(as described in the project) and newest versions of IDEA.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -37,6 +37,23 @@ patchPluginXml {
     untilBuild = ""
 }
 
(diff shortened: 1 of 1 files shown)


hadarshjfrog:
Shorten the PR name please - so it'll be clear in the release notes later

hadarshjfrog (approved):

PR #482, opened 2024-09-12
Fixed the title for warnings in webview

eyalk007:
-----
The title in the webview became the description. this obviously isn't good for anything as it is not the right pararmater.


eyalk007:
before fix
image

attiasas (approved):
LGTM

eyalk007:
image
original bug is still solved

reference here https://github.com/jfrog/jfrog-idea-plugin/pull/481

PR #483, opened 2024-09-17
Fixed dependency tree right click action

eyalk007:
-----

This PR fixes the right click action in the tree for dependencies with vulnerabilities.

Package json now works with scopes, and Maven and Gradle work now when inheriting version from parent. 

The problem for gradle and maven mainly resided in the fact we sent the version(that was not needed) to the comparison with the impact tree.
In package json,  the scope from the package name was removed and as a result, caused issues in identifying the package.


src/main/java/com/jfrog/ide/idea/navigation/NavigationService.java
```diff
@@ -49,8 +49,10 @@ public void addNavigation(DependencyNode treeNode, PsiElement navigationTargetEl
         }
         NavigationTarget navigationTarget = new NavigationTarget(navigationTargetElement, document.getLineNumber(navigationTargetElement.getTextOffset()), componentName);
         Set<NavigationTarget> navigationTargets = navigationMap.get(treeNode);
+
```
hadarshjfrog, line 52: all the changes are spaces? we can remove then...

src/main/java/com/jfrog/ide/idea/scan/GradleScanner.java
```diff
@@ -34,7 +34,7 @@
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.Map;
+import java.util.*;
```
hadarshjfrog, line 37: no change in the code, so why bring all of java.util?

src/main/java/com/jfrog/ide/idea/ui/LocalComponentsTree.java
```diff
@@ -108,24 +109,28 @@ private void handleContextMenu(ComponentsTree tree, MouseEvent e) {
...
         if (selected instanceof DependencyNode) {
-            createNodePopupMenu((DependencyNode) selected);
+            DescriptorFileTreeNode descriptorFileTreeNode = (DescriptorFileTreeNode) selectedPath.getParentPath().getLastPathComponent();
+            String descriptorPath =  descriptorFileTreeNode.getSubtitle();
+            this.createNodePopupMenu((DependencyNode) selected, descriptorPath);
         } else if (selected instanceof VulnerabilityNode) {
             createIgnoreRuleOption((VulnerabilityNode) selected, e);
         } else if (selected instanceof ApplicableIssueNode) {
             createIgnoreRuleOption(((ApplicableIssueNode) selected).getIssue(), e);
         } else {
```
hadarshjfrog, line 127: instead of removing the comment here - I suggest adding comments to all for cases of the if/else

src/main/java/com/jfrog/ide/idea/ui/LocalComponentsTree.java
```diff
@@ -108,24 +109,28 @@ private void handleContextMenu(ComponentsTree tree, MouseEvent e) {
...
         if (selectedPath == null) {
             return;
         }
         Object selected = selectedPath.getLastPathComponent();
+
         if (selected instanceof DependencyNode) {
-            createNodePopupMenu((DependencyNode) selected);
+            DescriptorFileTreeNode descriptorFileTreeNode = (DescriptorFileTreeNode) selectedPath.getParentPath().getLastPathComponent();
+            String descriptorPath =  descriptorFileTreeNode.getSubtitle();
+            this.createNodePopupMenu((DependencyNode) selected, descriptorPath);
```
hadarshjfrog, line 122: why is only this one under "this?

hadarshjfrog (commented):
The code changes themselves are ok - but from what I read in the description - we're eliminating all version comparison. 
How can that effect complex project with multiple packages from different types. should we identify them as vulnerabilities?

src/main/java/com/jfrog/ide/idea/navigation/NavigationService.java
eyalk007: um i dont see spaces in my code..

src/main/java/com/jfrog/ide/idea/scan/GradleScanner.java
eyalk007: not there anymore

src/main/java/com/jfrog/ide/idea/ui/LocalComponentsTree.java
eyalk007: okay tell me your opinion

eyalk007:
@hadarshjfrog. what your saying about the versions is a dilema. I can try and keep it in cases it exists(meaning from parent) and not in cases it does not.

but you will always have the possibility of a case where there are two parents in the project. one has version lets say 1.0.0
and the other has version 1.2.0

they both will look the same as they both dont have a version(because they both inherit from their parents)
a possible solution is to use the maven and gradle apis(as they are the only supported multi module projects). but we didn't want to rely on them from the beginning.

hadarshjfrog (approved):

PR #484, opened 2024-09-18
Fixed the description of SAST issues

eyalk007:
-----
Now we take the full description instead of the reason


hadarshjfrog:
Please add a screenshot of before and after the change - so it'll be clearer in the changelog

eyalk007:
after:
image

image

eyalk007:
before:

image


image

PR #487, opened 2024-10-27
Checking all CVEs in applicability scan

eyalk007:
-----

Unlike the `vscode extension` and the `cli`, the `IDEA plugin` does not run the applicability scanner on indirect cves.
This PR changes the behavior to be like the rest of our products.


eyalk007:
closed as support for contextual analysis scan for indirect will be added in the future

PR #493, opened 2025-01-26
Updated Readme About newly supported IDEs

eyalk007:
-----

Updated readme about supported IDEs (`Rider` and `CLion`)

### DIFF
--- a/src/main/resources/META-INF/plugin.xml
+++ b/src/main/resources/META-INF/plugin.xml
@@ -43,6 +43,7 @@
     <depends config-file="with-maven.xml" optional="true">org.jetbrains.idea.maven</depends>
     <depends config-file="with-go.xml" optional="true">org.jetbrains.plugins.go</depends>
     <depends config-file="with-python.xml" optional="true">com.intellij.modules.python</depends>
+    <depends optional="true" >com.jetbrains.rider</depends>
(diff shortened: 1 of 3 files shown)


src/main/resources/META-INF/plugin.xml
```diff
@@ -43,6 +43,7 @@
     <depends config-file="with-maven.xml" optional="true">org.jetbrains.idea.maven</depends>
     <depends config-file="with-go.xml" optional="true">org.jetbrains.plugins.go</depends>
     <depends config-file="with-python.xml" optional="true">com.intellij.modules.python</depends>
+    <depends optional="true" >com.jetbrains.rider</depends>
```
hadarshjfrog, line 46: doesn't rider has a config-file like the others?

eyalk007: if needed we will add.
For now it seems there is no need.

hadarshjfrog: Can we add a minimal one? for consistency sake?

eyalk007: i really dont think adding a stub there is necessary.
for other languages/packages this is needed for visiting the element(clicking on it and reaching the line)
 we did not add support for nugget so this is not needed

eyalk007:
closed as there is already an existing pr

PR #496, opened 2025-03-06
Changed the SAST title to the short description rule

eyalk007:
-----
Created a function to determine the reason according to scanner type, as when the scan is of SAST, we take the short description


eyalk007:
space changes is because i formatted the (wasnt done before)

attiasas (approved):

PR #498, opened 2025-03-25
Updated am version to 1.15.2

eyalk007:
-----

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -27,7 +27,7 @@ intellij {
 }
 
 runPluginVerifier {
-    verifierVersion = '1.383'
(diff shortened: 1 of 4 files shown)


hadarshjfrog (approved):

build.gradle
```diff
@@ -26,6 +26,10 @@ intellij {
     pluginName = 'JFrog'
 }
 
+runPluginVerifier {
```
attiasas, line 29: Please add a TODO, removing specific verifier version when fixed, include link to the bug

attiasas (approved):

PR #499, opened 2025-04-01
Project declaration itself triggers a violation from xray

eyalk007:
-----
Added handeling for cases where the project itself is a violation (for example rule on licenses that the project itself violates),
now when we group to descriptors, we check if the vulnerable dependency is part of the root of the descriptor.
if so, we add it appropriately.

### DIFF
--- a/src/main/java/com/jfrog/ide/idea/scan/SingleDescriptorScanner.java
+++ b/src/main/java/com/jfrog/ide/idea/scan/SingleDescriptorScanner.java
@@ -61,22 +61,21 @@ public abstract class SingleDescriptorScanner extends ScannerBase {
     protected List<FileTreeNode> groupDependenciesToDescriptorNodes(Collection<DependencyNode> depScanResults, DepTree depTree, Map<String, Set<String>> parents) {
         DescriptorFileTreeNode fileTreeNode = new DescriptorFileTreeNode(descriptorFilePath);
         for (DependencyNode dependency : depScanResults) {
-            boolean directDep = false;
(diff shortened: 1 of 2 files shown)


src/main/java/com/jfrog/ide/idea/scan/SingleDescriptorScanner.java
```diff
@@ -62,11 +62,16 @@ protected List<FileTreeNode> groupDependenciesToDescriptorNodes(Collection<Depen
...
-                    break;
+            if(dependency.getComponentIdWithoutPrefix().equals(depTree.rootId())) {
+                directDep = true;
+            }
+            else {
+                for (String parentId : parents.get(dependency.getComponentIdWithoutPrefix())) {
+                    DepTreeNode parent = depTree.nodes().get(parentId);
+                    if (descriptorFilePath.equals(parent.getDescriptorFilePath())) {
+                        directDep = true;
+                        break;
```
attiasas, line 73:
```suggestion
            if(dependency.getComponentIdWithoutPrefix().equals(depTree.rootId())) {
                dependency.setIndirect(false);
                continue
            }
            for (String parentId : parents.get(dependency.getComponentIdWithoutPrefix())) {
                    DepTreeNode parent = depTree.nodes().get(parentId);
                    if (descriptorFilePath.equals(parent.getDescriptorFilePath())) {
                        directDep = true;
                        break;
                
```

try to remove the `else`, keep code 'flat'

attiasas (commented):
LGTM, please add PR description and fix Title

attiasas (approved):
LGTM!

PR #501, opened 2025-05-07
Updated IDE common version to 2.3.7

eyalk007:
-----
Updated to a patch version that updates the go absolutizer and solves a bug in go that projects wiht a patch in the go version, cant be parsed by the absolutizer

also updated the python test with a new package


attiasas (approved):
LGTM

PR #512, opened 2025-09-16
Fix PypiScannerTest

kerenr-jfrog:
-----
- Changed an assert that was too tightly coupled to the exact number of direct and transitive dependencies, since that number can change over time.


src/test/java/com/jfrog/ide/idea/scan/PypiScannerTest.java
```diff
@@ -120,7 +120,6 @@ public void testBuildTree() throws IOException {
         // Check direct dependency
         String directDepId = DIRECT_DEPENDENCY_NAME + ":" + DIRECT_DEPENDENCY_VERSION;
         DepTreeNode scrappy = getAndAssertChild(results, results.getRootNode(), directDepId);
-        assertSize(16, scrappy.getChildren());
```
attiasas, line 123: I would add a test for the number of children for the root (direct count)

attiasas (commented):
I would replace the test and not delete it

PR #531, opened 2026-06-16
Fix GlobalSettings crash when JFrog CLI is not installed (XRAY-145646)

Jordanh1996:
## Summary

Fixes **XRAY-145646** — IntelliJ plugin 2.9.0 fails to start on Windows with:

```
PluginException: Cannot init component state (GlobalSettings)
Caused by: NullPointerException: Cannot invoke Log.error() because "this.log" is null
  at JfrogCliDriver.runVersion → isJfrogCliInstalled
```

**Root cause:** `ide-plugins-common` 2.4.x (#525) changed `JfrogCliDriver(Map, String path)` to `JfrogCliDriver(Map, Log)`. `ServerConfigImpl.readConnectionDetailsFromJfrogCli()` still passed `null` as the second argument — previously the CLI path, now the `Log`. When `jf` is not on PATH (common on Windows), `isJfrogCliInstalled()` calls `log.error()` and NPEs during `GlobalSettings` init.

**Fix:** Pass `new NullLog()` instead of `null` when constructing `JfrogCliDriver`.

## Test plan

- [x] Reproduced NPE on Windows VM without `jf` on PATH (broken 2.9.0 build)
- [x] Verified fixed build: IntelliJ starts, Global Configuration saves and persists across restart
- [x] `ConnectionDetailsFromCliTest` updated and passes

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/main/java/com/jfrog/ide/idea/configuration/ServerConfigImpl.java
+++ b/src/main/java/com/jfrog/ide/idea/configuration/ServerConfigImpl.java
@@ -34,6 +34,7 @@ import com.jfrog.ide.idea.ui.configuration.ConnectionRetriesSpinner;
 import com.jfrog.ide.idea.ui.configuration.ConnectionTimeoutSpinner;
 import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.lang3.Strings;
+import org.jfrog.build.api.util.NullLog;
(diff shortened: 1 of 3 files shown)


attiasas (approved):
LGTM, I would try to add test that could replicate the issue you fixed

